PR #183, opened 2014-08-08
server name indication - server-side support

hannesm:
do not merge yet - what is missing are examples for multiple hosts, which actually use SNI.
also, combinators to easen the configuration of `own_cert option * own_cert list` for the common use case(s)
just an initial set of changes to review (lwt/* should be ignored for now)


pqwy:
Ok, too bad you have a bunch of other PRs based on top of this so this is currently immutable.

As discussed, needs maor:
* `cert -> ciphersuites` structure to have validation a) in one place and b) at construction-time, so as to provide the handshake engine with cert-ciphersute combinations which it can assume make sense (and to make decisions in that part purely structural);
* name comparison semantics wants to live in a single, canonical place;
* algorithmics are broken, can spot n^2, I fear there are even darker corners.

I consider this WIP, to be revisited after everything that depends on this goes through.

But excellent starting point!

hannesm:
thanks! :dancers:

hannesm:
the dependency is https://github.com/mirleft/ocaml-tls/pull/192 (which I just rebased)

PR #191, opened 2014-09-22
AEAD / CCM

hannesm:
support for CCM cipher suites, this reworks some parts of TLS to get rid of the invariant that each ciphersuite has a hmac algorithm and needs a hmac_secret


hannesm:
I've rebased once more... and feel it is ok to merge..

PR #220, opened 2015-01-05
Server-side downgrade prevention

g2p:


g2p:
Here's a simple patch to implement TLS_FALLBACK_SCSV (#5).

It can be tested with

    openssl s_client -connect netloc -fallback_scsv #should succeed
    openssl s_client -connect netloc -fallback_scsv -no_tls1_2 #should fail

I couldn't integrate this with the openssl script in the testsuite though, because the test server seemed not to close things properly, causing EADDRINUSE the second time around. Maybe some tests of the handshake logic (high-level but without involving TCP) could be added later.

lib/handshake_server.ml
```diff
@@ -241,8 +245,9 @@ let answer_client_hello_reneg state (ch : client_hello) raw =
 
   let process_client_hello config oldsession oldversion ours ch =
     let cciphers = ch.ciphersuites in
+    let fallback = List.mem Packet.TLS_FALLBACK_SCSV cciphers in
     assure (client_hello_valid ch) >>= fun () ->
-    agreed_version config.protocol_versions ch.version >>= fun version ->
+    agreed_version config.protocol_versions ch.version fallback >>= fun version ->
```
g2p: I kept that line, but I don't think it's useful. The line that checks that version = oldversion is sufficient.

hannesm:
thanks for that contribution.. I've to read through it a bit more -- do you also plan to implement the client side?

g2p:
Probably not; I think on the client side it's better never to initiate a downgrade.
(browsers do it, but they deal with a lot of crap for the sake of stuff that used to work by luck)

hannesm:
sounds good to me!

hannesm:
patch looks good to me - I'd remove the processing of SCSV in renegotiation, though (since there is no defined semantics in the draft, and we currently do not allow version modifications during a renegotiation)

g2p:
Sure. Can I remove the `agreed_version` call during renegotiation? It has no effect since config.protocol_versions is static and the next line checks that ch.version is identical.

g2p:
Rebased and updated. I removed agreed_version from renegociation and added to the README's list of implemented drafts and RFCs.

g2p:
Right! I'm not used to monadic style and I'd missed that >>= reused the value.
While testing, I noticed (unrelated issue) that the alert version number doesn't match the version the client sends; it's always 3.3, probably the default state of the handshake.

pqwy:
--> #224

pqwy:
Also. @g2p -- thanks for the initiative here!

PR #223, opened 2015-01-12
do not advertise RC4

### LINKED ISSUE #8: RC4 security in TLS
problem: some bytes can be recovered

solution: throw RC4 away and switch to CBC

references
- http://www.isg.rhul.ac.uk/tls/index.html

hannesm:
disable RC4 by default (users who need RC4 can tweak the configuration)

fixes #8

### DIFF
--- a/lib/config.ml
+++ b/lib/config.ml
@@ -39,8 +39,6 @@ module Ciphers = struct
     `TLS_RSA_WITH_AES_256_CBC_SHA ;
     `TLS_RSA_WITH_AES_128_CBC_SHA ;
     `TLS_RSA_WITH_3DES_EDE_CBC_SHA ;
-    `TLS_RSA_WITH_RC4_128_SHA ;
-    `TLS_RSA_WITH_RC4_128_MD5
     ]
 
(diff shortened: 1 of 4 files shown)


lib/config.ml
```diff
@@ -39,8 +39,6 @@ module Ciphers = struct
     `TLS_RSA_WITH_AES_256_CBC_SHA ;
     `TLS_RSA_WITH_AES_128_CBC_SHA ;
     `TLS_RSA_WITH_3DES_EDE_CBC_SHA ;
-    `TLS_RSA_WITH_RC4_128_SHA ;
-    `TLS_RSA_WITH_RC4_128_MD5
     ]
```
pqwy: Well, this is a bit of a lie; removing ciphers we support from the list of supported ciphers is, um... yeah.

Nothing that couldn't be helped by a little rename and little rethink of what this list exactly means, or what potential similar lists defined next to it could mean.

pqwy:
Evil, evil RC4.

hannesm:
I agree we need multiple sets here - of what the protocol implementation implements and what we're happy to use by default..

PR #225, opened 2015-01-13
Added 'error_message' function

talex5:
This depends on changes to `FLOW` in https://github.com/mirage/mirage/pull/346.
It will allow TLS to be used with Conduit + Xen without losing errors.


pqwy:
Included and superseeded by #258.

PR #231, opened 2015-01-22
validate certificates when a certificate message is received, also

hannesm:
if the certificate list is empty.

this changes the behaviour slightly:
 - if authenticator is provided in config, fail when no certificate is
   sent (fails in Handshake_common.parse / Certificate.parse_stack)


hannesm:
(yeah, travis is supposed to fail since it depends on https://github.com/mirleft/ocaml-x509/pull/37)

lib/handshake_common.ml
```diff
@@ -248,21 +243,27 @@ let validate_chain authenticator certificates hostname keytype usage =
...
-    ((s, xs), Some anchor)
+    authenticate authenticator hostname certs >>= fun anchor ->
+    key_size Config.min_rsa_key_size certs >>= fun () ->
+    ( match certs with
+      | s :: _ ->
+        guard (validate_keytype s keytype &&
+               validate_usage s usage &&
+               validate_ext_usage s `Server_auth)
+          Packet.BAD_CERTIFICATE
+      | _ -> return () ) >|= fun () ->
```
hannesm: I've the feeling this `match certs with` and keytype and usage checks need to be moved to handshake_client, since the validation of client certs will most likely be different!? (currently for client auth, hostname is `None`, keytype `RSA` and usage `Digital_signature`)

PR #236, opened 2015-01-26
Add `with sexp` to `client` and `server`

pqwy:
No more harm than having non-quite-bijective serialization for `config` that's already there.


hannesm:
needs to be in config.mli as well, the `with sexp`...

pqwy:
Sorry, somehow managed not to commit.

PR #240, opened 2015-02-02
support io-page 1.3.0+

talex5:
Needed to support the latest version of io-page.


pqwy:
Ah, I just pushed https://github.com/mirleft/ocaml-tls/pull/241 before seeing this.

@talex5 Cross-check my thinking, please: aligning is not needed after https://github.com/mirage/mirage-net-xen/commit/7d2777d789530e2104b6d9252cdca0d6f8078acb, right?

talex5:
Yeah, that's a better fix! `FLOW` says the buffer need not be aligned, so `tls` is doing the right thing (technically `tcpip` is doing the wrong thing by sending unaligned data to `NETWORK`, but mirage-net-xen now handles that).

PR #248, opened 2015-02-08
remove duplicated traces

hannesm:
we can still reproduce the internal states:
 - the outgoing state (state-out) is always recorded
 - the initial state not recorded (it is known at the beginning, and the same as the last outgoing state)
 - handshake-in/handshake-out is not recorded (the raw packets are recorded and thus these can be reproduced)
--> tracing of handshake packets was primarily useful for visualization in the demonstration server, but not for trace checking

the amount of producde data is reduced by roughly 50%...

### DIFF
--- a/lib/engine.ml
+++ b/lib/engine.ml
@@ -360,7 +360,7 @@ let assemble_records (version : tls_version) : record list -> Cstruct.t =
 (* main entry point *)
 let handle_tls state buf =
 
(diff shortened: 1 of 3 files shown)


hannesm:
maybe we should be able to trigger those traces optionally from the outside, to allow the demonstration server to work (again in some time, once its bitrot is cleaned up). i'm happy to get this merged in its current state with commented out tracing..

PR #250, opened 2015-02-08
Yet more sexps

pqwy:
Gets rid of the local cstruct pretty-printing code, delegates to the more compact native version.

Merging should probably wait for https://github.com/mirage/ocaml-cstruct/pull/44, but this is what Piñata runs right now.


lib/state.ml
```diff
@@ -43,42 +43,36 @@ type cipher_st =
...
 
-let crypto_context_of_sexp _ = failwith "can't parse crypto context from sexp"
-and sexp_of_crypto_context cc =
-  Sexp_ext.record [
-    "sequence" , sexp_of_int64 cc.sequence ;
-    "cipher_st", sexp_of_cipher_st cc.cipher_st
-  ]
-
+let cipher_st_of_sexp =
+  Conv.of_sexp_error "cipher_st_of_sexp: not implemented"
```
avsm: if this is an error, why not just expose only `sexp_of` below?

pqwy: One, to get downstream types to use `with sexp` and avoid the boilerplate of writing [one side of] it by hand. Two, because this is expected to be fixed eventually (it's in a `(* Sexplib stubs *) ... (* *** *)` comment block), and it's nice to have the API, even if sketched out.

Note that all these types should be invisible to the outside. This is all basically for tracing/debugging.

hannesm:
now that cstruct-1.6.0 is out, could you rebase? then we could merge and be happy..

avsm:
And update the opam file constraint at same time as rebase

pqwy:
Rebased.

But once merged, it will require the unreleased `mirage-xen 2.3.2` with [support](https://github.com/mirage/mirage-platform/commit/3d343935f451ba6c505e5008d2e458cf506bad8d) for `cstruct 1.6.0`.

ping @avsm

hannesm:
thanks!

PR #254, opened 2015-02-15
Fix Lwt entropy

pqwy:
Now with less stupid: does not force the event loop to periodically wake up.


lwt/tls_lwt.ml
```diff
@@ -18,6 +18,8 @@ let resolve host service =
   | ai::_ -> return ai.ai_addr
 
 
+module Unix_ = Unix
+
```
avsm: that's pretty ugly.  You only use this for `gettimeofday`, so why not replace this with `let gettimeofday = Unix.gettimeofday` ?

pqwy: Yeah, that was pretty ugly.

PR #256, opened 2015-03-08
always require secure renegotiation

hannesm:
not requiring this leaves us otherwise open to MITM prefix
attacks.  RFC5746 is out since > 5 years.  no reason to allow
the insecure option.


hannesm:
within 30 000 server side traces 900 clients had no secure renegotiation... partially due to broken ubuntu/openssl setups I believe... happy to drop these clients (similar to SSLv3)...

pqwy:
Ab-so-lu-te-ly.

PR #257, opened 2015-03-09
adjust to new X509 API: cert_type returns option, cert_pubkey is present

hannesm:
depends on https://github.com/mirleft/ocaml-x509/pull/43


hannesm:
this and the corresponding x509 PR are good to merge!

PR #264, opened 2015-04-20
update extension_type enum

davegarrett:
https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml

Update TLS extension ID enum and RFC/draft comments.

This brings the list here up-to-date with the current IANA registry. This just adds the values; it doesn't actually do anything with them. The /rfc dir needs updating with RFC text as well.

### DIFF
--- a/lib/packet.ml
+++ b/lib/packet.ml
@@ -101,7 +101,7 @@ cenum compression_method {
   LZS     = 64;
 } as uint8_t (sexp)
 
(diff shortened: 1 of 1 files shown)


hannesm:
thank you very much for this update! we need to develop parsers, and high-level representations for the missing extensions (and then also implement their logic).

PR #269, opened 2015-05-13
adapt to X509 API changes in signing branch

hannesm:
this goes with https://github.com/mirleft/ocaml-x509/pull/48


pqwy:
Could you remove the x509 pin part, please, so that it can be cleanly merged?

hannesm:
done

PR #270, opened 2015-05-14
rename pfs to fs -- there's nothing perfect in forward secrecy

hannesm:
(actually, a better name is `key erasure` (djb))

### DIFF
--- a/lib/config.mli
+++ b/lib/config.mli
@@ -93,16 +93,18 @@ module Ciphers : sig
   (** {1 Cipher selection} *)
 
   val default : ciphersuite list
-  (** All ciphersuites this library uses by default. *)
+  (** [default] is a list of ciphersuites this library uses by default. *)
 
   val supported : ciphersuite list
-  (** All ciphersuites this library implements. *)
+  (** [supported] is a list of ciphersuites this library supports
+      (larger than [default]). *)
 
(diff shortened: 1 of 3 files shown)


lib/config.mli
```diff
@@ -98,11 +98,11 @@ module Ciphers : sig
...
 
-  val pfs : ciphersuite list
-  (** All perfect forward secrecy ciphersuites this library supports. *)
+  val fs : ciphersuite list
+  (** All forward secrecy ciphersuites this library supports. *)
 
-  val pfs_of : ciphersuite list -> ciphersuite list
-  (** [pfs_of ciphers] selects all perfect forward secrecy ciphersuites from default. *)
+  val fs_of : ciphersuite list -> ciphersuite list
+  (** [pfs_of ciphers] selects all forward secrecy ciphersuites from default. *)
```
samoht: need to rename the in the doc as well

hannesm: thanks, done!

pqwy:
Well, Wikipedia redirects from [Perfect forward secrecy](en.wikipedia.org/wiki/Perfect_forward_secrecy) to [Forward secrecy](http://en.wikipedia.org/wiki/Forward_secrecy), so I guess that settles the naming dispute.

PR #283, opened 2015-06-19
Session resumption

hannesm:
This adds support for session resumption. The user-facing change is that both `Config.client` and `Config.client` have a callback parameter `session_cache : Cstruct.t option -> epoch_data option`. A client uses the session id from `session_cache None` in its initial client hello. A server resumes a session if the id is found in the cache.

The cache may be filled by a user by using the `epoch_data` (exposed from `Engine.epoch : state -> epoch_data`) and inserting it into a cache, using its `sessionid` as key.

Internal cleanups: `epoch_data` moved to `Core` module; `Handshake_crypto` has `derive_master_secret` and `initialise_crypto_context` separated. Bot the client and the server state machine have two new states, waiting for CCS/Finished from the other side during resumption. Both `session_data` and `epoch_data` now contain a `sessionid` field. `sever_hello` is a top-level function in `Handshake_server`, which produces a server hello (analogue to `default_client_hello`).

While doing a renegotiation we do not do resumption: if someone request new symmetric key material, reusing the same and just running the KDF is useless (since if master_secret is compromised, all input to it (server_random, client_random transmitted over the wire) is known.

### DIFF
--- a/lib/handshake_client.ml
+++ b/lib/handshake_client.ml
@@ -26,11 +26,15 @@ let default_client_hello config =
     match version with
     | TLS_1_0 | TLS_1_1 -> List.filter (o not Ciphersuite.ciphersuite_tls12_only) cs
     | TLS_1_2           -> cs
+  and sessionid =
+    match config.session_cache None with
+    | None -> None
+    | Some { session_id ; _ } -> Some session_id
   in
   let ch = {
     version      = Supported version ;
     random       = Rng.generate 32 ;
-    sessionid    = None ;
+    sessionid    = sessionid ;
     ciphersuites = List.map Ciphersuite.ciphersuite_to_any_ciphersuite ciphers ;
     extensions   = host @ signature_algos
   }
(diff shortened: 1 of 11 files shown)


lib/handshake_client.ml
```diff
@@ -26,11 +26,15 @@ let default_client_hello config =
     match version with
     | TLS_1_0 | TLS_1_1 -> List.filter (o not Ciphersuite.ciphersuite_tls12_only) cs
     | TLS_1_2           -> cs
+  and sessionid =
+    match config.session_cache (Cstruct.create 0) with
```
avsm: This `Cstruct.create 0` is a bit of an odd interface.  Is it mutated or realloced or something by the `session_cache` function?

lib/handshake_client.ml
```diff
@@ -59,21 +63,42 @@ let answer_server_hello state ch (sh : server_hello) raw log =
...
+      Handshake_crypto.initialise_crypto_ctx state.protocol_version session
+    in
+    let machina = AwaitServerChangeCipherSpecResume (session, client_ctx, server_ctx, log @ [raw]) in
+    ({ state with protocol_version = sh.version ; machina = Client machina }, [])
+  | _ ->
+    let machina =
+      let cipher = sh.ciphersuites in
+      let session_id = match sh.sessionid with None -> Cstruct.create 0 | Some x -> x in
+      let session =
+        { empty_session with
```
avsm: inconsistent formatting to the `let session` binding above (line 70)

hannesm:
thx, I thought a bit more about session_cache, and rewrote history: `session_cache : Cstruct.t option -> epoch_data option`

PR #287, opened 2015-06-30
Session hash

hannesm:
based on resumption #283 an implementation of session hash (proposed draft). resumption is only possible if an extended_master_secret was used (due to triple handshake the only sensible way to deal with it)

### DIFF
--- a/lib/core.ml
+++ b/lib/core.ml
@@ -82,6 +82,7 @@ type extension =
   | Padding of int
   | SignatureAlgorithms of (Hash.hash * signature_algorithm_type) list
   | UnknownExtension of (int * Cstruct.t)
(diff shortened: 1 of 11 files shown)


hannesm:
interop (of extended master secret) was tested successfully with polarssl

hannesm:
@pqwy would be great if you could read through #283 and this PR...

hannesm:
rebased to master. This change integrates session hash (RFC 7627), and only proceeds with resuming a session if the old and new session support session hash, to prevent triple handshake mitm attacks. might be a bit too conservative for the use case where someone runs a public web server and wants to get speed improvements from resumption; there enabling resumption if renegotiation is off should be secure.

PR #292, opened 2015-07-02
Implied entropy for Lwt

pqwy:
Since the previous design was untenable, I made sure entropy init is actually completed *synchronously* ([docs](https://github.com/mirleft/ocaml-nocrypto/blob/master/lwt/nocrypto_entropy_lwt.mli#L10)). This means that we can **FINALLY** internalize entropy init and people can just forget about the whole thing.


hannesm:
yay! :dancer:

PR #297, opened 2015-09-09
expose own_random and peer_random in epoch_data

cfcs:
Expose `client_random` and `server_random` from `lib/state.ml` to allow extracting the authenticated timestamp from a remote party:

```ocaml
type session_data = {
  server_random    : Cstruct.t ; (* 32 bytes random from the server hello *)
  client_random    : Cstruct.t ; (* 32 bytes random from the client hello *)
  (* [...] *)
}
```

in `epoch_data` as `own_random` and `peer_random`, see `lib/engine.mli` ->
```ocaml
val epoch : (state : State.state) -> (epoch : epoch_data)`

### DIFF
--- a/lib/core.ml
+++ b/lib/core.ml
@@ -163,9 +163,11 @@ type master_secret = Cstruct.t with sexp
 type epoch_data = {
   protocol_version : tls_version ;
   ciphersuite      : Ciphersuite.ciphersuite ;
+  peer_random      : Cstruct.t ;
   peer_certificate : X509.t list ;
   peer_name        : string option ;
   trust_anchor     : X509.t option ;
+  own_random       : Cstruct.t ;
   own_certificate  : X509.t list ;
   own_private_key  : Nocrypto.Rsa.priv option ;
   own_name         : string option ;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by cfcs** [comment]
Note: This doesn't check the state of the connection, so for the timestamp use case above, you'd want to check that the connection has been established, see `lib/engine.ml`:
```ocaml
let handshake_in_progress (s: State.state) = (* ... *)
```

- something like:

```ocaml
let is_connection_established s =
  match s.handshake.machina with
  | Server Established
  | Client Established -> true
  | _                  -> false
```

hannesm:
thanks!
`epoch` returns the `InitialEpoch` variant (without any epoch_data) until a TLS connection is established, thus there is no need for any checks... the caller of `epoch` needs to handle the possibility that there is no data yet (on a not established session)..

cfcs:
Ah, great! I was a bit unsure when it made the shift, but that makes sense.

hannesm:
@pqwy I'm happy  to merge this unless you have any objections

pqwy:
Nope, it's ok by me.

We do have a bit of record-filed-creep though. Longer term, if significantly more stuff gets added to the epoch [and related structs], I'd think of structuring it and potentially even making it abstract (w/ accessors).

But it's cool so far.

PR #300, opened 2015-10-15
by default, switch off renegotiation.  common use case does not need renegotiation

hannesm:


hannesm:
reasoning for this is a) the common use cases, b) reading the discussion https://mailarchive.ietf.org/arch/search/?email_list=tls&gbt=1&q=clarification+on+interleaving

hannesm:
to elaborate a bit more: since triple handshake it seems to be unsafe to use both resumption and renegotiation at the same time (unless session hash is used).  until recently we didn't support resumption.

now that we have resumption, the question is what is more common: renegotiation or resumption (without session hash).  it is pretty obvious that most use cases include resumption, rather than renegotiation (apart from some "weird" client-authentication-which-do-not-reveal-client-identity (such as tor's second handshake method)).  thus changing the default to no renegotiation (which opens the resumption field (if a store is provides)) looks reasonable.

PR #305, opened 2015-12-03
do not hardcode path to certificates and private keys ("tls/"), fixes #294

### LINKED ISSUE #294: The path to keys is hardcoded
It is possible to change the name of the keys, but not the path, which is hardcoded to `tls/`.

The most confusing is that, this is not clear at all when building from the pieces available at https://github.com/mirage/mirage-seal/tree/master/static

hannesm:
this is certainly a breaking API change and all earlier mirage applications
will break, but forcing a "tls/" path in the store is imho bad.

### DIFF
--- a/mirage/example/sekrit/ca-roots.crt
+++ b/mirage/example/sekrit/ca-roots.crt
@@ -1,0 +1,1 @@
+../../../certificates/ca-root-nss-short.crt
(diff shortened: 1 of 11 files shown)


yomimono:
This user welcomes the breaking API change.

avsm:
Agreed, a very sensible change.

samoht:
yay!

PR #306, opened 2015-12-04
allow resumption if either no renegotiation OR extended master secret

hannesm:
previously, this was only allowed if the extended master secret was provided
(followup of #300, which includes reasoning)


hannesm:
the second commit fixes client-based resumption...

PR #308, opened 2015-12-04
X.509 updates for path building:

hannesm:
- returned failures are different
- preserve more in the epoch_data:
 - peer_certificate (t option)
 - received_certificates (t list)
 - peer_certificate_chain (t list)
 - trust_anchor (t option)


hannesm:
previously the peer certificate was extracted from the list of received certificates at various occasions.  now, this lives in a field in the session data.  the verified trust chain is also explicitly saved, next to the raw list of received certificates.

PR #318, opened 2016-02-17
Type equality

pqwy:
`Or_error` is generative. Distinct functor applications generate distinct types. This induces juggling between internally generated modules. In addition, it got used in the parsing code just for the error type, not in its monad capacity.

Introduces `Result`, a type that will be builtin in 4.03; moves parsers to pure `Result`, to dis-entangle them from the core effect type; and reuses `Result` as the error type in `Or_error`. Various instances of`Ok | Error` are now mutually compatible.

This is a stepping stone to fully redefine the underlying effect as in the other [branch](https://github.com/mirleft/ocaml-tls/tree/nondeterministic-tls), but is worth reconciling with TLS1.3 work and merging in its own right.


hannesm:
(just rebased and pushed tls13 on top of master with this PR merged)

PR #320, opened 2016-02-27
get rid of camlp4, use ppx_sexp_conv, cstruct.ppx, and no lwt.syntax anymore

hannesm:


rgrinberg:
Friendly ping. Would be great if the tls stack was released without camlp4, I'd be able to get rid of camlp4 in conduit then.

avsm:
@hannes got time to finish this one today? Would unblock the web stack updates. Even a point release with the use of `pa_sexp_conv` instead of `sexplib.syntax` would help move past the constraints

PR #321, opened 2016-03-11
use functoria style for mirage example

azul:
I kept the second example the way it was. It seems to not be too flexible anyway in terms of target architecture.

### DIFF
--- a/mirage/example/config.ml
+++ b/mirage/example/config.ml
@@ -9,37 +9,15 @@ let build =
     | "server" -> `Server
   with Not_found -> `Server
 
-let disk =
-  match get_mode () with
-  | `Unix -> direct_kv_ro secrets_dir
-  | `Xen  -> crunch secrets_dir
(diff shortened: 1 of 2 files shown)


hannesm:
thanks!

mirage/example/config.ml
```diff
@@ -9,37 +9,15 @@ let build =
...
-    | "dhcp"   -> `Dhcp
-    | "static" -> `Static
-  with Not_found -> `Static
-
-let stack console =
-  match net, dhcp with
-  | `Direct, `Dhcp   -> direct_stackv4_with_dhcp console tap0
-  | `Direct, `Static -> direct_stackv4_with_default_ipv4 console tap0
-  | `Socket, _       -> socket_stackv4 console [Ipaddr.V4.any]
+let stack console = generic_stackv4 console tap0
```
hannesm: this actually prevents me now from running it on unix with the socket backend...

PR #328, opened 2016-07-20
Implement V1_LWT.FLOW disconnect

djs55:
`FLOW.close` should flush data and signal to the remote that no more data will be written.

`FLOW.disconnect` should disconnect from the flow immediately without signalling and possibly drop in-flight/in-buffer data.

`FLOW.close` should act like `Unix.shutdown` while `FLOW.disconnect` should act like `Unix.close`.

This patch implements `FLOW.disconnect` by setting any `Active` connection to `Eof` (hopefully the GC will deallocate it later) and calling `disconnect` on the underlying flow.

Related to [mirage/mirage#550]

Signed-off-by: David Scott dave@recoil.org

### DIFF
--- a/mirage/tls_mirage.ml
+++ b/mirage/tls_mirage.ml
@@ -156,6 +156,13 @@ module Make (F : V1_LWT.FLOW) = struct
       FLOW.(write flow.flow buf >>= fun _ -> close flow.flow)
     | _           -> return_unit
 
+  let disconnect flow =
+    ( match flow.state with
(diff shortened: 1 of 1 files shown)


pqwy:
Is this still relevant?

LGTM

hannesm:
it seems relevant, but let's wait till https://github.com/mirage/mirage/pull/550 is merged

hannesm:
thanks a lot, this has been superseeded by #488

PR #329, opened 2016-08-17
Use PCLOCK / Ptime

mattgray:

PR #331, opened 2016-08-22
Use PCLOCK / Ptime

hannesm:
I massaged the commits of #329 slightly: no need for `ptime` in `tls-lwt` right now, and add a dependency onto `ptime >= 0.8.1` (for `Ptime.v`) into the opam file.  This is only needed for the mirage subpackage, but we're not able to express this in opam (we could explicit create multiple packages (might be worth doing, but then we should also switch to `topkg`), or provide a new `mirage-no-mirage` (similar to `mirage-no-xen`/`mirage-no-solo5`) package (feels way to hacky)).

Travis CI will not succeed with mirage since it uses mirage as released into opam-repository (and I don't think there's much value in moving this towards mirage-dev branches).  Likely the example unikernels suffer from bitrot atm, but can be fixed in subsequent commits.

@mattgray I squashed your commits and amended the changes mentioned above.  hope that's fine with you.

### DIFF
--- a/_oasis
+++ b/_oasis
@@ -55,8 +55,8 @@ Library "tls-mirage"
   FindlibParent   : tls
   Path            : mirage
   Modules         : Tls_mirage
-  BuildDepends    : x509, tls, mirage-types, ipaddr, lwt
-  XMETARequires   : x509, tls, mirage-types, ipaddr, lwt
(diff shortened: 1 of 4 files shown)


mattgray:
fine with me @hannesm :)

PR #332, opened 2016-09-25
Implement server acceptable certificate auths

reynir:
The code for accepted certificate authorities in a client certificate
request was already implemented, but handshake_server.ml had hardcoded
an empty list of CAs. This commit adds a configuration option for this.

Note that this requires a change to ocaml-x509: https://github.com/mirleft/ocaml-x509/pull/87

### DIFF
--- a/lib/config.ml
+++ b/lib/config.ml
@@ -28,6 +28,7 @@ type config = {
   authenticator     : X509.Authenticator.a option ;
   peer_name         : string option ;
   own_certificates  : own_cert ;
+  accepted_cas      : X509.t list ;
   session_cache     : session_cache ;
   cached_session    : epoch_data option ;
 } [@@deriving sexp]
@@ -91,6 +92,7 @@ let default_config = {
   authenticator     = None ;
   peer_name         = None ;
   own_certificates  = `None ;
+  accepted_cas      = [] ;
   session_cache     = (fun _ -> None) ;
   cached_session    = None ;
 }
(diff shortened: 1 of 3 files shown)


reynir:
I should mention that I haven't tested this thoroughly. I have tested with `openssl s_client`. As mentioned, the TLS-part seem to have been implemented and it was just a matter of passing CAs and extracting the issuer DN (hence the ocaml-x509 change). It seems to work - I have tested with one and two certificates in `config.accepted_cas`.

The **TODO** in lib/handshake_server.ml mentions extracting "DN from authenticator", but since authenticator is a function it wouldn't be possible so I opted for adding Yet Another Configuration Option accepted_cas. It's not the best user experience since you have to repeat the CAs, once in the authenticator and once in the server config. Furthermore, the list of CAs will be static - there might be some use case for a dynamic list of CAs(?)

hannesm:
thanks. indeed, I'd prefer to have the `authenticator` and `accepted_cas` unified.  given that authenticator might as well be a set of fingerprints, this is certainly not easily doable atm, but we'll for sure find a way.  I'm currently travelling..

reynir:
Done. Have a nice nap+trip! :-)

cfcs:
Did this ever get merged?

reynir:
@cfcs No, it's not been merged.

hannesm:
thanks, superseeded by #374 (please let me know if you disagree with the changes in 374, I amended your commit, i.e. you're still the author :)

reynir:
Thanks!

The logic seems to be the same modulo naming and DN-unwrapping, so no complaints from me.

I look forward to play with client certificates again!

PR #335, opened 2016-10-17
Update certificate source to the AOSP bundle

### LINKED ISSUE #334: Update certificate source
Currently this is a bit messy.

There are three files:
- ca-root-nss.crt
- ca-root-nss-short.crt
- cacert.crt

CACert shouldn't be trusted, it has never been accepted by Mozilla and its purpose has been obsoleted by Let's Encrypt.

ca-root-nss.crt is a very old extraction from Mozilla's trusted set, and contains compromised authorities like CNNIC.
ca-root-nss-short.crt is the above, minus the readable version of the certificates, plus cacert.crt.

I recommend using the ca-certificates bundle from Android AOSP, which is a subset (about 9 removals so far) of Mozilla's certdata.txt reference bundle. The only minor downside (for the git log) is that the order will change.

It doesn't sound optimal to duplicate the list (otherwise everything is loaded twice), so I'd suggest keeping the readable version for easier maintenance. The shorter version can be extracted with sed: https://stackoverflow.com/questions/28895208/

g2p:
This is a subset of Mozilla's; see
https://android-developers.blogspot.com/2016/07/changes-to-trusted-certificate.html

Closes #334.


pqwy:
Undecided.

This merely changes debug material.

hannesm:
@pqwy agreed!

this PR is an improvement, and we should have some library which ships trust anchors: on unix/win32 use the system anchors (similar to what haskell is doing in https://github.com/vincenthz/hs-certificate), for freestanding systems we should have a script which imports FF or Chrome trust anchors (thus freeing ourselves from maintaining a CA list) -- I'll likely code this up soon.

another item is to not ship the certificates with ocaml-tls while installing, since they're purely for debugging/testing

hannesm:
the whole `certificates` subdirectory is only used for debugging and testing, and not installed during build+install.  I'll close this PR, and agree that maybe there's need for a trust anchor library in OCaml.

PR #336, opened 2016-10-21
don't require a peer name for client_of_fd, there's no need for this

hannesm:
This is the low-level interface, and not all TLS scenarios require any form of name validation.

An example is a syslog server, we don't want to depend on DNS for syslog.  Authentication is done via a CA chain, but no name is compared.  See https://github.com/hannesm/logs-syslog/blob/32137a34e83e5a0ba6ba8c6e8da45957971de71e/src/logs_syslog_lwt.ml#L65

### DIFF
--- a/lwt/tls_lwt.ml
+++ b/lwt/tls_lwt.ml
@@ -187,8 +187,11 @@ module Unix = struct
       tracer = trace ;
     }
 
-  let client_of_fd ?trace config ~host fd =
-    let config'     = Tls.Config.peer config host in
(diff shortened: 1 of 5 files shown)


hannesm:
the second commit adjusts the mirage interface in a similar fashion, making `?host` an optional argument and set the `peer_name` if a `host` is provided.  Both interfaces could be more minimal: drop the `?host`, since `peer_name` is part of the configuration -- but there's no way to construct a config with `peer_name <> None` (atm!).  Having the host explicit is nice IMHO.

PR #337, opened 2016-10-24
Topkg

hannesm:
this got slightly bigger than expected, I also removed the pack (and provided a main interface, tls.mli).  compilation with `principal` fails (thus not included in `_tags`), and there are various warnings which I'll fix subsequently


hannesm:
@pqwy if you feel like, pls review+merge

pqwy:
Why is this so ginormous?

hannesm:
I removed `pack`, put a `tls.mli` together by copy&pasting, and prefixed the modules with `Tls_`.

pqwy:
Yes. But _why_?

hannesm:
- remove `pack`: since there's no reason for `pack` (and I had the impression it is in general a good property to not `pack` libraries) -- unlikely that OCaml will be able to remove modules during link time (since Engine basically needs all the modules)
- rename, prefix with `Tls_`: since in current master if you try `make doc`, it will fail due to the module `Config` being defined in `Tls` and elsewhere.  `pack` seems to work fine in some scenarios to hide the internal name, but not always (and OCaml lacks namespacing).  again, I thought this is best practises
- `Tls.mli` to provide an API, a single file where potential clients of this package can dig into

pqwy:
`pack` is a pretty neutral thing, orthogonal to other concerns.

Except if you remove it, hide other modules, and slap a common `mli` on top, it's a game changer. Previously, it used to be the case that "we didn't have a clear separation between the public and private API" -- as both of us repeatedly wrote -- because the parts saw each other through the same interfaces that the clients saw them through. Since other modules now don't go through `tls.mli`, it is precisely this: canonical public interface.

So this PR piggy-backs an extremely important thing, introducing an explicit public API where there was none, onto a set of changes ostensibly about the build system switch. For reasons that are beyond me.

And it does so in the worst possible way. Look at the new, explicitly public, [interface](https://github.com/mirleft/ocaml-tls/blob/1580929e5baff0ba04cd56547cb10582d3be0ca0/lib/tls.mli). It's leaking the internals left and right. Of course it is, it's a **concatenation of private APIs**.

It made me chuckle a bit. I seem to remember a published paper about how the moral equivalent of that file is in fact short and simple.

So could you please cut down the PR to deal with changing the actual build system, while producing the same artifact. And then we can talk about what's the final public footprint of the entire library.

hannesm:
I agree that the public API needs more thought (IMHO we can do this incrementally on master, but you seem to disagree).  I started in #338 with fixes for travis (and moving the examples to mirage-dev).

PR #342, opened 2016-11-13
another topkg attempt

hannesm:
instead of in #337, I focus only on the topkg bits, not any interface design


hannesm:
managed to solve (hopefully) myself by adding a `api:[]` into `pkg.ml` (and using path 2 above).

hannesm:
(update: no, can't get it working)

hannesm:
update: works now, pls look whether you're happy with this and if so, merge. thx.

pqwy:
Could you just pls squash a stylesheet removal into that? `topkg` will use `odig`'s stylesheet, so no need for every repo to have its own snapshot of it.

Blue links. **shudder**

hannesm:
done

PR #345, opened 2016-12-03
do not crash on double close

vbmithr:
Prevents failure when the user calls `Lwt_io.close` on both `ic` and `oc`.


vbmithr:
@hannesm @pqwy ping?

vbmithr:
It happens when the user call close on `ic` and `oc`. It is very easy to forget that closing one will automatically close the other since they share the same file descriptor.

hannesm:
thanks!

PR #354, opened 2017-01-03
fix broken link

c-cube:


hannesm:
thanks for this PR.  I uploaded current documentation (using topkg) to the doc url mentioned in README.

PR #358, opened 2017-01-28
Update .merlin

rgrinberg:
* syntax extensions are no longer used
* add lwt.unix to the list of findlib packages


hannesm:
thanks, merged manually

PR #364, opened 2017-03-06
Remove X509_lwt.extension

rgrinberg:
This function is already available in the stdlib. The only difference
is that it returns `""` when there's no extension.


rgrinberg:
This function  is actually 4.04+ at this point

PR #365, opened 2017-06-29
opam updates, unbreak (client side) renegotiation with EMS, authentication

hannesm:
My motivation behind these changes is to allow the scenario where the client should authenticate, but transmit its certificate only over an encrypted connection.  The way to do this is (from the server side, since the server has to choose that it needs an authenticated client) to establish a (only server-authenticated) TLS connection, switch the authenticator in place (to send out a CertificateRequest in the next handshake), and then ask the client to renegotiate the connection (by sending a HelloRequest).

While implementing, I found an issue (since 0.7.0 / summer 2015) which avoided our client code from renegotiating if the ExtendedMasterSecret extension is used (fixed in a4783c5).

The design of `Tls_lwt.reneg` was far from finished, esp. with recent changes (23 Jan 2016) to `can_handle_appdata` (and its usage in `Tls_lwt.reneg` around 1 Jul 2014) lead that it returned too early (before the renegotiation was finished).  The commit bb4186a reintroduces a `handshake_in_progress` which is used in `drain_handshake` for looping.

The semantics of `reneg` needed another tweak, since application data is allowed between the HelloRequest and the ClientHello, what to do with it.  There are two choices: retain or drop.  Drop is the new default, and an optional parameter is introduced.  This allows server code to reason about the epoch application data arrived -- or to ensure within server code that data received in the last epoch is discarded.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ script: bash -ex .travis-opam.sh
 env:
   global:
     - PACKAGE="tls"
-    - PINS="x509"
+    - PINS="x509:https://github.com/hannesm/ocaml-x509.git#newcrl"
   matrix:
     - OCAML_VERSION=4.02 DEPOPTS=lwt
     - OCAML_VERSION=4.03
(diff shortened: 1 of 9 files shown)


hannesm:
apart from the renegotiation commits, this has been upstreamed.  I'll work on the renegotation today and open a new PR.

PR #373, opened 2017-12-16
Change

hannesm:
these are the bugfixes excluding renegotiation from #365 including x509 0.6.0 compatibility


hannesm:
the CI failure is addressed in https://github.com/ocaml/opam-repository/pull/11048

PR #378, opened 2018-03-09
Support ALPN Extension

bobbypriambodo:
An attempt to implement ALPN Extension ([RFC7301](https://tools.ietf.org/html/rfc7301)).

I've arranged the change into standalone commits to (hopefully) ease review.

What's currently implemented: reading & writing ALPN extension, no_application_protocol alert, configuring broadcasted/preferred protocols via config (+ validation), protocol selection (+ validation), and appending the selected protocol into `handshake_state.alpn_protocol`.

However, there are some questions that I'm not sure I know the answer to:

1. How should users retrieve the selected protocol from the state, i.e. how should the public API be? I imagine I can add a `get_alpn_protocol : state -> alpn_protocol` function on `Tls.Engine`, but I don't know how to use it from `Tls_lwt` and `Tls_mirage` (I think those two are the main entry points people use, cmiiw).

2. In some implementations (e.g. Go's `crypto/tls` library), instead of throwing `no_application_protocol` alert on no matching client/server protocol [as per the RFC](https://tools.ietf.org/html/rfc7301#section-3.2), they chose to not fail and instead return a field noting whether a mutual protocol has been negotiated, I think it's so that clients can decide by themselves what to do. Openssl 1.0.2n I think also does this, but iirc the latest version fails according to the RFC. Should we stick with the RFC implementation?

3. From the RFC I'm not sure whether specific TLS versions need specific treatment, so I just assume that this extension will work on all TLS versions. If this assumption is wrong, then I would need to add further checks.

4. It seems that `certificates/server.pem` has expired, so the examples at `lwt/examples` mostly failed with certificate expired errors. Do you have any pointers on how to generate a fresh cert file?

Thanks.


bobbypriambodo:
I'm not sure what caused the build error, it's for mirage builds.

```
ocamlfind: Package `tcp_socket_options' not found - required by `tcpip.tcpv4-socket'
```

Does anyone have an idea?

rgrinberg:
@yomimono fixes this in mirage-tcpip master I believe. Can we have a release perhaps?

yomimono:
Yes, the pull request was only just merged.  I'll cut a release today, if uplink allows.

yomimono:
tcpip version 3.4.1 release PR is at https://github.com/ocaml/opam-repository/pull/11541 .

bobbypriambodo:
Thanks @rgrinberg @yomimono!

hannesm:
@bobbypriambodo thanks for your pull request, sorry for the delay in replying. to discuss your questions:
1. `tls` already provides session information via [epoch](http://mirleft.github.io/ocaml-tls/doc/Engine.html#VALepoch), which is exposed in [tls_lwt](http://mirleft.github.io/ocaml-tls/doc/Tls_lwt.Unix.html#VALepoch) and [tls_mirage](http://mirleft.github.io/ocaml-tls/doc/Tls_mirage.Make.html#VALepoch).  This should be enriched (in module `Core`) with a `peer_alpn` field in the record (likely of type .
2. depending on the use case and what we see in the wild, I'm usually in favour to stick with the RFC and wait until someone needs to continue even without alpn agreement
3. don't know, haven't read the RFC yet
4. the opam package `certify` exposes X509 CA functionality via command line utilities (`selfsign`, `csr`, `sign`), feel free to add a new certificate and private key to your PR.

Codewise I have some questions.  I suspect, again the client sends a list of alpn possibilities and the server replies with which one is used in this session?
- how is preference taken into account?  for the cipher list, the server has a configured sorted sequence, where the first one matching one of the ciphersuites proposed by the clients is selected (i.e. the server configured list is taking preference using its order). -- looking into the tests you added, none exercises the preference afaict (since the order you use is always the same)
- how does alpn work with renegotiation?  is this specified at all?  can i establish a session without alpn and later renegotiate and ask for a alpn?  does the alpn of handshake n need to be the same as the alpn of session n+1?

bobbypriambodo:
Thanks for the reply! I'll try to answer your questions:

> the client sends a list of alpn possibilities and the server replies with which one is used in this session?

Yes. The ClientHello includes a ProtocolNameList field which contains a "list of protocols advertised by the client, in descending order of preference." The server replies with the same field structure, "except that the ProtocolNameList MUST contain exactly one ProtocolName".

> how is preference taken into account?

As the above, the client advertises in descending order of preference. As for the server selection, I'm putting the exact RFC text here:

```
It is expected that a server will have a list of protocols that it
supports, in preference order, and will only select a protocol if the
client supports it.  In that case, the server SHOULD select the most
highly preferred protocol that it supports and that is also
advertised by the client.  In the event that the server supports no
protocols that the client advertises, then the server SHALL respond
with a fatal "no_application_protocol" alert.
```

Following this spec, both `Config.client` and `Config.server` functions accepts a list of preferred protocols in a descending order. The code that does the selection is the following:

https://github.com/bobbypriambodo/ocaml-tls/blob/77ce5c8959af4632c1c7a7625fa97bbe6d75ac25/lib/handshake_server.ml#L340-L352

I represent `selected_protocol` as an `alpn_protocol list option` due to the need of failing in case both client and server advertises ALPN but have no matching protocol. It's a bit dirty, but I'm not sure how to properly handle it otherwise.

I will add tests to simulate the preference selection.

> how does alpn work with renegotiation? is this specified at all? can i establish a session without alpn and later renegotiate and ask for a alpn? does the alpn of handshake n need to be the same as the alpn of session n+1?

From what I understand from the RFC, you can establish a session without alpn and later renegotiate and ask for an alpn, and alpn of handshake n does not need to be the same as session n+1. Copying here the sections I deemed relevant:

```
Unlike many other TLS extensions, this extension does not establish
properties of the session, only of the connection.  When session
resumption or session tickets [RFC5077] are used, the previous
contents of this extension are irrelevant, and only the values in the
new handshake messages are considered.
```

and

```
The protocol identified in the
"application_layer_protocol_negotiation" extension type in the
ServerHello SHALL be definitive for the connection, until
renegotiated.  The server SHALL NOT respond with a selected protocol
and subsequently use a different protocol for application data
exchange.
```

> tls already provides session information via epoch

This is where I'm confused, due to the sentence "this extension does not establish
properties of the session, only of the connection", I was in doubt whether it is proper to put the information inside `epoch`. But seeing it again, it would seem that `epoch` is really the way to go?

> if you add another commit replacing 216.34.181.45 by 216.105.38.15 in mirage/example/unikernel.ml, that should work better ;)

Okay, I'll do just that. Thanks!

bobbypriambodo:
Hi, I have added more commits to this PR based on the previous feedback. Sorry for the long wait.

1. I've added a piece of [docs on ALPN protocol selection](https://github.com/mirleft/ocaml-tls/pull/378/commits/1b86f8bb8187db41b96e489bd1cf6154385d18f6). Hopefully this may give clarity on how ocaml-tls implementation will pick the protocol. I put it on `config.mli` since it made the most sense to me at this time, but I can move it if there's a more proper place.

2. [Added `echo_server_alpn.ml`](https://github.com/mirleft/ocaml-tls/pull/378/commits/0ce76e17ec50d26a4b32c6790bb04350ad3db3b7) (broken on < 4.04.0, fixed by [this one](https://github.com/mirleft/ocaml-tls/pull/378/commits/3bbdbabe71c5e7219b3bd97ba9aaef9955258696)) to manually test the protocol selection implementation (I'm still not sure how to add automated tests for this, would love a few pointers), e.g.:

    ```
    $ _build/lwt/examples/echo_server_alpn.native foo,bar
    [server] -> start @ 4433 (use `openssl s_client -connect host:4433 -alpn <proto>`), available 
    protocols: foo,bar

    # in another terminal
    $ openssl s_client -connect 127.0.0.1:4433 -alpn bar,foo
    ...
    Server public key is 1024 bit
    Secure Renegotiation IS supported
    ALPN protocol: foo
    ...
    ```

3. Added [`peer_alpn` to `epoch_data`](https://github.com/mirleft/ocaml-tls/pull/378/commits/1495cfbeb09d4a01408d55d9a4fa0fc7469b928c). Again, I'm a bit unsure to test this as I'm still figuring out how to properly extract epoch data from client. It seems I would have to use the low-level `Tls_lwt.Unix` API?

hannesm:
thanks @bobbypriambodo, I'll hopefully find time this wednesday to review.

hannesm:
I re-evaluated and came up with https://github.com/bobbypriambodo/ocaml-tls/pull/1

bobbypriambodo:
Hi @hannesm, if you could have a look at it again, I just added an example of retrieving the selected protocol with `Tls_lwt`.

To test it, you can do:

```bash
$ ./build
$ openssl s_server -cert certificates/server.pem -key certificates/server.key -alpn h2,foo

# on different terminal session

$ _build/lwt/examples/echo_client_alpn.native
...trace output...

Selected protocol: h2
```

bobbypriambodo:
After reading more about this, some implementations like OpenSSL provide a more flexible control of protocol selection, in that they allow users to attach a callback ([SSL_CTX_set_alpn_select_cb](https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_set_alpn_select_cb.html)), which receives the list of protocols from the client and expects a protocol in return. That means users can freely use any selection algorithm on that callback.

Our implementation here strictly favors the sorted protocol list given by the server, in that it gets the first server protocol that is also supported by the client regardless of client preference. IMO this still adheres to the RFC, but let me know if you want to use the callback approach instead.

hannesm:
Dear @bobbypriambodo, thanks for that information.  I currently don't have much time to review this PR again - what needs to be evaluated is test coverage in respect to the new code.  I think the approach is fine - to configure ALPN once when the `config` is built (and would not be in favour of any callback mechanism).

Besides being busy with other things, I've no use for ALPN, thus moving slowly here - please bear with me.

bobbypriambodo:
Oh, sure, I'm really glad you're looking into this. I'm sorry if I came across as a bit pushy, because I really want to see HTTP/2 support in OCaml and the lack of ALPN support in both ocaml-tls and ocaml-ssl (which I'm also currently working on) seems to be one of the core blockers.

Would you have any idea on who might have an interest on this PR that I could ask for help? (Or I could ask in OCaml Discourse if you don't mind.)

Noted on the coverage, I'll see what I can do about it.

hannesm:
> I really want to see HTTP/2 support in OCaml

same here, but isn't another missing piece a HTTP/2 protocol implementation?

> Would you have any idea on who might have an interest on this PR that I could ask for help?

hmm, what kind of help are you looking for?  I'll hopefully find some time soon to merge this PR, but before I'd like to review the diff once more and convince myself that all edge cases are covered, and the information is stored at the right places.

> Noted on the coverage, I'll see what I can do about it.

I guess since now the server hello may only contain a single ALPN protocol, the failing case (which errors out early in `Parse`) should be part of the test suite (if this is not already the case, can't remember).

bobbypriambodo:
> isn't another missing piece a HTTP/2 protocol implementation?

Yeah, that too. I imagine it would also need a nontrivial work. Not sure if I'm up to the task, but if there's no one working on it yet then perhaps I can take a look after this.

hannesm:
thanks, squashed and merged in #382

PR #380, opened 2018-06-22
Add handhake test client to examples

bramford:
I found that many TLS-enabled [web]sites fail silently when attempting a TLS handshake as a client.

A common issue is a mismatch of the ciphers. This client tests all ciphers in parallel and notes the cause of any failures.

### DIFF
--- a/lwt/examples/.merlin
+++ b/lwt/examples/.merlin
@@ -1,0 +1,3 @@
+PKG tls lwt lwt.ppx
+S ./
+B _build/
--- a/lwt/examples/handshake_test_client.ml
+++ b/lwt/examples/handshake_test_client.ml
@@ -1,0 +1,76 @@
+(* Build with: ocamlbuild -use-ocamlfind -pkgs lwt,lwt.ppx,lwt.unix,tls,tls.lwt handshake_test_client.native *)
+let connect host port ca_dir cipher =
+  let%lwt authenticator = X509_lwt.authenticator (`Ca_dir ca_dir) in
+  try%lwt
+    let%lwt ic, oc =
+      Tls_lwt.connect_ext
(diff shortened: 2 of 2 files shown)


cfcs:
@bramford It doesn't seem to be handling network errors?

bramford:
@cfcs Do you have an example? Feel free to add any extra error conditions to the `try%lwt ... with` match. I'll test and add more when I get the chance.

bramford:
I've updated it to catch Unix errors. I imagine there are other possible errors but I haven't seen them.

hannesm:
thanks for your pull request.  would you mind to include this example client into the package build system (by adjusting `_tags` and `pkg/pkg.ml`)?  It is a bit unclear to me whether this program should be part of the ocaml-tls repository, or rather hosted (and distributed) as standalone repository.  as it is now, no test programs are actually installed by ocaml-tls (but there may be use for `openssl s_client` and `openssl s_server` applications).

bramford:
@hannesm Am I correct that you're suggesting a new project to provide CLI-based TLS client & server testing application(s) similar to `openssl s_client` & `openssl s_server`? I agree that this makes more sense than building these tools in to this repo. However, in order to test _ocaml-tls_, one would have to know or assume that such a tool exists outside the _ocaml-tls_ repo. Unlike with openssl where you get the test client & server tools with the rest of the runtime.

cfcs:
Where do you see this fitting into that vision @hannesm? https://github.com/hannesm/tlsclient

hannesm:
closing, we can always add more tests (there's already test_client, echo_client, http_client) -- but please with a rebased approach on the main branch, and using cmdliner (plus dune as build system).

PR #384, opened 2018-10-05
use new Mirage_kv.RO interface

samoht:
Implement the new interface from https://github.com/mirage/mirage-kv/pull/14


hannesm:
included as #392 thanks

PR #387, opened 2018-12-27
Don't try writing to the remote if we got `EOF`

### LINKED ISSUE #347: handle shutdowns and last message sent by server gracefully

anmonteiro:
This should fix #347 and is an alternative to #371.

I was receiving EPIPE signals in my code that caused the last chunk sent
by the server to never get in my application.

After some investigation, I realized that, in `Tls_lwt.Unix.read_react`,
the library tries to write a response (if there's one) on the socket.
However, in the case that the state is EOF, writing this response will
throw SIGPIPE and never deliver the received data to the calling code.

The proposed fix is to not even try to write this response in case we
got EOF from the socket.

### DIFF
--- a/lwt/tls_lwt.ml
+++ b/lwt/tls_lwt.ml
@@ -82,7 +82,10 @@ module Unix = struct
             | `Alert a -> `Error (Tls_alert a)
           in
           t.state <- state' ;
-          (resp |> when_some (write_t t)) >>= fun () -> return (`Ok data)
+          if state' = `Eof then
+            return (`Ok data)
+          else
+            (resp |> when_some (write_t t)) >>= fun () -> return (`Ok data)
 
       | `Fail (alert, `Response resp) ->
           t.state <- `Error (Tls_failure alert) ;


rymdhund:
Yes, I can confirm that this patch fixes the crash I had in https://github.com/mirleft/ocaml-tls/issues/347#issuecomment-348719725

anmonteiro:
@hannesm thanks for the review. perhaps guarding against the write failing and always returning the data is a better approach.

I reproduced this bug with Tls_lwt as a client.

The conditions where I managed to repro this bug were:
- after the tls handshake, send and HTTP request and shutdown the underlying socket for sending
- because TLS may need to write on the socket even after sending the request (this is my naive understanding), and it was shutdown, we got EPIPE.

hannesm:
@anmonteiro do you have some example code of your TLS client usage?
> send and HTTP request and shutdown the underlying socket

you shouldn't `shutdown` the underlying socket ;)

> TLS may need to write on the socket even after sending the request (this is my naive understanding)

Yes, the teardown process of a TLS connection consists of one side sending the alert `close_notify`, the other side replies with a `close_notify` (as mentioned elsewhere, a peer doesn't have to wait for the `close_notify` from the other side before `close`ing the socket.

anmonteiro:
@hannesm it's not easy to extract my code, but basically using `Tls_lwt` without [this http/af patch](https://github.com/inhabitedtype/httpaf/pull/84) is what was causing the issue for me.

hannesm:
thanks, tls 0.9.3 is released (PRed to opam-repository https://github.com/ocaml/opam-repository/pull/13231) which fixes this issue

PR #391, opened 2019-02-25
fix extension length if padding is injected into client hello

hannesm:
the extension length rewritten used to be wrong, by including the 2 byte extension length field (which needs to be excluded). it is likely nobody noticed since in <= TLS 1.2 it's rare that padding is injected


hannesm:
I tested this manually with openssl s_server on the other side (and stumbled upon it only when doing some TLS_1_3 handshakes where this code path is executed all the time)

PR #392, opened 2019-02-25
bugfixes and release prep

hannesm:
- the extension length rewritten used to be wrong, by including the 2 byte extension length field (which needs to be excluded). it is likely nobody noticed since in <= TLS 1.2 it's rare that padding is injected
- use mirage-kv 2.0.0 interface in tls-mirage (see https://github.com/mirage/mirage-kv/pull/14, taken from #384)


hannesm:
CI failure is due to mirage (and mirage-types etc) need a version bump, which in exchange requires a tls release. to solve, I'll merge and release this, and restart travis once mirage 3.5.0 is released.

PR #395, opened 2019-08-14
Lwt - only close descriptor once

hcarty:
Before this change cohttp-lwt-unix + conduit-lwt-unix + ocaml-tls would end up with double-close errors.  For example, `cohttp-curl-lwt https://www.google.com/ -v -o /dev/null` with tls 0.10.3 logs `EBADF` when the connection is closed.  That no longer occurs with this change applied.

Somewhat related to https://github.com/mirage/ocaml-conduit/pull/294

### DIFF
--- a/lwt/tls_lwt.ml
+++ b/lwt/tls_lwt.ml
@@ -237,8 +237,13 @@ type ic = Lwt_io.input_channel
 type oc = Lwt_io.output_channel
 
 let of_t ?close t =
+  let only_if_open f =
+    let is_open = ref true in
+    (* Only close the underlying fd once *)
+    fun () -> if !is_open then (is_open := false; f ()) else Lwt.return_unit
+  in
   let close = match close with
-    | None   -> (fun () -> Unix.(safely (close t)))
+    | None   -> (only_if_open (fun () -> Unix.(safely (close t))))
     | Some f -> (fun () -> Unix.safely (f ()))
   in
   (Lwt_io.make ~close ~mode:Lwt_io.Input (Unix.read_bytes t)),


hcarty:
cc @avsm

hannesm:
thanks for your PR, though I'm not entirely sure I can follow your changes. The API `Tls_lwt.of_t` already provides an optional named argument `~close`, thus if the caller wants to keep the file descriptor open (or close it manually), you could provide `~close:(fun () -> Lwt.return_unit)`. Wouldn't that work for your scenario? (I'm sorry I don't really know and understand cohttp-lwt-unix / conduit-lwt-unix)

hcarty:
The default `close` used in `Tls_lwt.of_t` will close the underlying file descriptor twice in its current state - once when closing the input channel, once when closing the output channel.  This PR ensures that the descriptor is only closed once even if the returned input and output channels are both closed.

This double close can be worked around in conduit/cohttp but the default in ocaml-tls probably shouldn't lead to a double close.

hcarty:
For a bit more clarity - I only want to change the default, and I _do_ want the default to continue to close the file descriptor when the input and output channels are closed.  I just don't want closing both to cause a double close on the file descriptor.

hannesm:
ok, wouldn't it then be sufficient to pass the `~close` only to either the input or output channel when calling `Lwt_io.make ~close ~mode:..`?

hannesm:
i.e. could you have a look if #396 would work for you? (I'm kind of afraid of leaking fds, that's why it would be great to have another pair of eyes on this) -- this avoids introduction of mutable state..

hcarty:
Only attaching a `close` to one channel would work as long as the caller always closes both channels. But if they forget to close the one which triggers the actual descriptor close then their code will leak descriptors. And I definitely share your fear of descriptor leaks!

hannesm:
hmm... back to square one, what is the actual issue with a double close (guarded inside of `Unix.safely`, which is `Lwt.catch (close) (fun _ -> Lwt.return_unit)`)? -- i.e. the potential raised exceptions from Lwt are caught and ignored.

hcarty:
My understanding, based on speaking with others who know more about this than I do, is that under higher concurrency levels file descriptor values can be recycled quickly enough that a double close could end up closing a separate, still living connection.

Outside of that danger, the EBADF errors are being seen and logged by cohttp somehow, which surprised me.

avsm:
If you double close an fd, then the underlying file descriptor number might be recycled and therefore accidentally closed.  Ideally we should only cause `close` once.  Ended up just tracking this within the channel structure in conduit... https://github.com/mirage/ocaml-conduit/pull/294/files#diff-7cb62769efef1a71cd4eb0ec32d355feR171

hcarty:
@avsm I was going to take that approach at first, but I don't know if a half-closed TLS connection makes sense - I'm not familiar enough with the standard to know.  Do you or @hannesm have any thoughts/suggestions there?

hannesm:
my thought is: why would you track the state of a file descriptor? shouldn't we instead use `match Lwt_unix.state fd with Closed -> Lwt.return_unit | _ -> Lwt.close fd`? if you're concerned that other stakeholders^Wlayers close the file descriptor, adding some mutable state on your layer won't avoid double closes at all, no? please see #397 for what I suggest

half-closed sessions: an endpoint signals "I won't send any more data" by sending a close_notify alert, and it already may cancel the read on the connection, thus half-closed connections are barely useful (in contrast to half-closed TCP sessions) -- see https://tools.ietf.org/html/rfc5246#section-7.2.1

avsm:
Makes sense! I think in conduit-lwt-unix, we wanted to do a `shutdown` if only one side was closed and then a full close once the second half finished.  We could probably use `Lwt_unix.state` there too to avoid the double tracking of state.

PR #397, opened 2019-08-15
tls-lwt: no double close

hannesm:
by first checking whether the fd has been closed already or not

### DIFF
--- a/lwt/tls_lwt.ml
+++ b/lwt/tls_lwt.ml
@@ -238,8 +238,12 @@ type oc = Lwt_io.output_channel
 
 let of_t ?close t =
   let close = match close with
-    | None   -> (fun () -> Unix.(safely (close t)))
     | Some f -> (fun () -> Unix.safely (f ()))
+    | None   -> (fun () ->
+        (* avoid double-closes by checking if the fd has already been closed *)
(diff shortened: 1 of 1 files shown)


hcarty:
This seems like a much cleaner approach vs #395!

cfcs:
Nice!

At the moment `mk_ch` (called by e.g. of_unix_file_descr`) assumes that he fd is open, so we can still end up in invalid states like this (unrelated to this PR):

```ocaml
# let fd = Unix.stderr in
Unix.close fd ;;
- : unit = ()
# Lwt_unix.state (Lwt_unix.of_unix_file_descr fd);;
- : Lwt_unix.state = Lwt_unix.Opened
```

hcarty:
This does fix the original test I was using under #395 (`cohttp-curl-lwt https://www.google.com/ -v -o /dev/null`).

hannesm:
@cfcs oh of course, the `Lwt_unix.state` is only lwt-internally tracked, i.e. if you use the `Unix` module for the underlying actual file descriptors, it won't work. This is something which may be worth discussing in the lwt issue tracker (whether there is a potential solution in lwt), but I don't think we will solve this in ocaml-tls.

cfcs:
Woops, wrong repo :)

Issue #399, opened 2019-10-10
Revokation checks are not enabled


emillon:
Hi,

This is just a heads-up that at the moment there's some support for checking revocation status in `ocaml-x509` but this is not wired all in:
- CRLs: they are not passed to the authenticator function (and there is no fetching & caching at the moment)
- OCSP: mirleft/ocaml-x509#34
- OCSP stapling: mirleft/ocaml-x509#35

Probably a mix of several of these will be necessary, unfortunately.

Thanks!

hannesm:
Thanks for your report. While OCSP is still not implemented, CRL are passed and checked in X.509 authenticators.

PR #404, opened 2020-01-22
X509.0.9.0: hash_whitelist and expose crl support

hannesm:
addresses some points of #399 and adapts to the new x509 API. on top of #403

### DIFF
--- a/lib/config.ml
+++ b/lib/config.ml
@@ -170,13 +170,14 @@ let non_overlapping cs =
   in
   let rec check = function
     | []    -> ()
-    | s::ss -> if not (List.for_all
-                         (fun ss' -> Domain_name.Set.is_empty (Domain_name.Set.inter s ss'))
-                         ss)
-               then
-                 invalid_arg "overlapping names in certificates"
-               else
-                 check ss
+    | s::ss ->
+      if not (List.for_all (fun ss' ->
+          X509.Certificate.Host_set.is_empty (X509.Certificate.Host_set.inter s ss'))
+          ss)
+      then
+        invalid_arg "overlapping names in certificates"
+      else
+        check ss
   in
   check namessets
 
--- a/mirage/tls_mirage.mli
+++ b/mirage/tls_mirage.mli
@@ -56,10 +56,15 @@ end
 
 (** X.509 handling given a key value store and a clock *)
 module X509 (KV : Mirage_kv.RO) (C : Mirage_clock.PCLOCK) : sig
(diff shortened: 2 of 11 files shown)


lib/config.ml
```diff
@@ -170,13 +170,14 @@ let non_overlapping cs =
...
-               else
-                 check ss
+    | s::ss ->
+      if not (List.for_all (fun ss' ->
+          X509.Certificate.Host_set.is_empty (X509.Certificate.Host_set.inter s ss'))
+          ss)
+      then
+        invalid_arg "overlapping names in certificates"
+      else
+        check ss
```
cfcs: Not that my suggestion below is particularly better, but a recursive function here doesn't seem necessary?

```ocaml
(* module M = X509.Certificate.Host_set *)
let check ss =
  let ss_count, unique =
    (* count each ss element, and add it to a union of all ss elements.*)
    List.fold_left (fun (count,acc) next ->
      count + M.cardinal next , M.union acc next)
      (0, M.empty) ss
  in
  if ss_count <> M.cardinal unique (* if not, duplicates in ss were merged in the union *)
    then invalid_arg "overlapping names in certificates"
in
check namessets
```

mirage/tls_mirage.mli
```diff
@@ -56,10 +56,15 @@ end
...
 (** X.509 handling given a key value store and a clock *)
 module X509 (KV : Mirage_kv.RO) (C : Mirage_clock.PCLOCK) : sig
-  (** [authenticator store clock typ] creates an [authenticator], either
-      using the given certificate authorities in the [store] or
-      null. *)
-  val authenticator : KV.t -> [< `Noop | `CAs ] -> X509.Authenticator.t Lwt.t
+  (** [authenticator ~hash_whitelist ~crl store typ] creates an [authenticator],
+      either using the given certificate authorities in the [store] as
+      value for key "ca_roots.crt", or null. If [hash_whitelist] is provided,
+      only these hashes are allowed for signatures of the certificate chain.
```
cfcs:
```suggestion
      only these hash algorithms are allowed for signatures of the certificate chain.
```
just to remove any shred of doubt that it could be a list of hash values similar to the `Key_fingerprints` argument to X509_lwt.authenticator

lib/config.ml
hannesm: if you feel strongly about that, please submit a separate PR with your code. I don't find either version particularly more readable -- and will stick to the smaller diff

PR #405, opened 2020-01-27
tls 1.3 support

hannesm:
needs ~~rebase +~~ review ~~+ get it compiling with released packages (fiat, ..)~~


lib/ciphersuite.ml
```diff
@@ -37,22 +38,31 @@ type aead_cipher =
...
+type payload_protection13 = [ `AEAD of aead_cipher ] [@@deriving sexp]
+
+type payload_protection = [
+  payload_protection13
+  | `Stream of stream_cipher * Nocrypto.Hash.hash
+  | `Block of block_cipher * Nocrypto.Hash.hash
+  ] [@@deriving sexp]
 
 (** [key_length iv payload_protection] is [(key size, IV size, mac size)] where key IV, and mac sizes are the required bytes for the given [payload_protection] *)
 let key_length iv pp =
```
cfcs: As discussed it would be nice to have a comment here explaining that this is only used for TLS `<= 1.2 `

lib/ciphersuite.ml
```diff
@@ -37,22 +38,31 @@ type aead_cipher =
   | AES_256_GCM
   [@@deriving sexp]
 
-type payload_protection =
-  | Stream of stream_cipher * Nocrypto.Hash.hash
-  | Block of block_cipher * Nocrypto.Hash.hash
-  | AEAD of aead_cipher
-  [@@deriving sexp]
+(* this is K_LEN, max 8 N_MIN from RFC5116 -- as defined in TLS1.3 spec *)
```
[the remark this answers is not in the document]
cfcs: (and would it make sense to mark this as `1.3`-only like `payload_protection13` below?)
(( wondering if it might be clearer to put the 1.3-functionality in a module here so it becomes more clear what is used where))

lib/config.ml
```diff
@@ -81,46 +107,72 @@ module Ciphers = struct
...
-       invalid "TLS 1.2 configured but no hashes provided"
-    | hs when not (List_set.subset hs supported_hashes) ->
-       invalid "Some hash algorithms are not supported"
-    | _                                                 ->
-       () ) ;
+  ( match config.signature_algorithms with
+    | [] when v_max >= TLS_1_2 ->
+       invalid "TLS 1.2 configured but no signature algorithms provided"
+    | hs when not (List_set.subset hs supported_signature_algorithms) ->
+       invalid "Some signature algorithms are not supported"
```
cfcs: As a user it would probably be nice to be told *which* algorithms are not supported (and the TLS version) so you can remove them from your config?

lib/ciphersuite.ml
hannesm: NB the IV for AEAD stems from https://tools.ietf.org/html/rfc5288#section-3 (for GCM) -- `salt[4]` (CCM https://tools.ietf.org/html/rfc6655#section-3)

lib/config.ml
```diff
@@ -31,27 +30,56 @@ module DN = struct
   let sexp_of_t _ = Sexplib.Sexp.Atom "distinguished name"
 end
 
+type ticket_cache = {
+  lookup : Cstruct.t -> (psk13 * epoch_data) option ;
+  ticket_granted : psk13 -> epoch_data -> unit ;
+  lifetime : int32 ;
+  timestamp : unit -> Ptime.t
+}
```
cfcs: It seems to me this structure heavily suggests using mutable state (with callbacks taking care of the mutation). Would it make sense to support a monadic approach like sketched below (where the mutable versions can just have `unit ticket_cache`?
```ocaml
type 'a ticket_cache = {
  cache_state : 'a ;
  lookup : 'a -> Cstruct.t -> (psk13 * epoch_data) option * 'a ;
  ticket_granted : 'a -> psk13 -> epoch_data -> 'a ;
  lifetime : int32 ; (* TTL seconds or absolute timestamp ? *)
  timestamp : 'a -> unit -> Ptime.t ;
}
(* alternatively I guess [ticket_cache] could be a module signature and the user would have to provide
    a module. *)
```
I'm not sure what your policy is, but I think it would be nice to eventually be able to serialize/deserialize the entire TLS state. Both for debugging/tracing/introspection and for applications that need to transfer it (supporting for instance session hand-off). I know that it's not immediately possible at the moment, but what do you think about that as an eventual goal?

hannesm: I prefer feature requests as separate issues.

lib/config.ml
```diff
@@ -67,8 +95,6 @@ module Ciphers = struct
     ]
 
   let supported = default @ [
-    `TLS_DHE_RSA_WITH_AES_256_GCM_SHA384 ;
-    `TLS_DHE_RSA_WITH_AES_128_GCM_SHA256 ;
     `TLS_RSA_WITH_AES_256_GCM_SHA384 ;
     `TLS_RSA_WITH_AES_128_GCM_SHA256 ;
     `TLS_DHE_RSA_WITH_3DES_EDE_CBC_SHA ;
```
cfcs: FWIW here we are announcing RC4 support, which the TLS 1.3 spec explicitly forbids (again, if we do not have a use-case for RC4 it would be sensible to strip support entirely to prevent mistakes):
https://www.rfc-editor.org/rfc/rfc8446.html#appendix-D.5
>    The security of RC4 cipher suites is considered insufficient for the
>   reasons cited in [RFC7465].  Implementations MUST NOT offer or
>   negotiate RC4 cipher suites for any version of TLS for any reason.

lib/config.ml
```diff
@@ -81,46 +107,72 @@ module Ciphers = struct
...
+    (* `TLS_AES_128_CCM_8_SHA256 *)
+  ]
+
+  let supported13 = default13
+
 end
 
-let default_hashes =
-  [ `SHA512 ; `SHA384 ; `SHA256 ; `SHA224 ; `SHA1 ]
+(* TODO split into <=12 and >=13, the SHA1 isn't 13 anymore *)
```
cfcs: Relevant entry in the spec seems to be in 4.2.3, page 44:
https://www.rfc-editor.org/rfc/rfc8446.html#page-44
(They differentiate between PKCS1+SHA1 used in certificates vs in the handshake)

lib/crypto.ml
```diff
@@ -73,6 +76,23 @@ let sequence_buf seq =
...
+    let l = Cstruct.len nonce in
+    let s = sequence_buf seq in
+    let pad = Cs.create (l - 8) in
+    pad <+> s
+  in
+  Cs.xor nonce s
+
+let adata_1_3 len =
+  let buf = Cstruct.create 5 in
+  Cstruct.set_uint8 buf 0 (Packet.content_type_to_int Packet.APPLICATION_DATA) ;
```
cfcs: I had to dig around a bit to understand what the two `\x03\x03` here are (the TLS_1_3 version tuple); it looks to me like this function is implementing https://www.rfc-editor.org/rfc/rfc8446.html#section-5.2 (page 81), which would be good to have documented?

This function seems very similar to `pseudo_header` (which implements https://tools.ietf.org/html/rfc5246#section-6.2.3.3 (top of page 25)) - that would also be good to have documented!

lib/writer.ml
```diff
@@ -23,7 +23,8 @@ let assemble_any_protocol_version version =
 
 let assemble_hdr version (content_type, payload) =
   let buf = create 5 in
-  set_uint8 buf 0 (content_type_to_int content_type);
+  let ct = content_type_to_int content_type in
+  set_uint8 buf 0 ct;
```
cfcs: Is this change a remnant from debugging?

lib/writer.ml
```diff
@@ -154,58 +181,152 @@ let assemble_alpn_protocol p =
...
+  assemble_ext @@ match e with
+    | `SupportedGroups groups ->
+      (assemble_supported_groups groups, SUPPORTED_GROUPS)
     | `Hostname name -> (assemble_hostnames [name], SERVER_NAME)
     | `Padding x ->
-       let buf = create x in
-       memset buf 0 ;
-       (buf, PADDING)
+      let buf = create x in
+      memset buf 0 ;
```
cfcs: `Cstruct.create` already fills with null-bytes, as opposed to `Cstruct.create_unsafe`. Maybe this changed upstream since this was written? In any case, the `memset` here is redundant with newer versions of `Cstruct` (perhaps we could just constrain `tls` to require `>=` such version)?

Conversely we also have a lot of `Cstruct.create 1 ; Cstruct.set_uint8 b 0 ...` that could maybe be optimized a bit; not sure if the scary look of `Cstruct.create_unsafe` is worth it, and I suspect the allocations themselves are much more expensive than the redundant memsetting at the moment.

lib/writer.ml
```diff
@@ -154,58 +181,152 @@ let assemble_alpn_protocol p =
...
+let assemble_retry_extension e =
+  assemble_ext @@ match e with
+    | `SelectedGroup g -> (assemble_group g, KEY_SHARE)
+    | `Cookie c -> (assemble_cookie c, COOKIE)
+    | `SelectedVersion v -> (assemble_protocol_version v, SUPPORTED_VERSIONS)
+    | `UnknownExtension _ -> invalid_arg "unknown retry extension"
+
+let assemble_cert_ext (certificate, extensions) =
+  let cert = assemble_certificate certificate
+  and exts = assemble_list Two assemble_server_extension (* ?? *) extensions
```
cfcs: What does `(* ?? *)` mean here?

lib/writer.ml
```diff
@@ -154,58 +181,152 @@ let assemble_alpn_protocol p =
...
+
+let assemble_certificates_1_3 context certs =
+  let l = create 1 in
+  set_uint8 l 0 (len context) ;
+  l <+> context <+> assemble_certs_exts (List.map (fun c -> c, []) certs)
+
+let assemble_sid sid =
+  let buf = create 1 in
+  match sid with
+  | None   -> set_uint8 buf 0 0; buf
```
cfcs: Same here re: `Cstruct.create` initializing with zeroes
```suggestion
  | None   -> buf
```

lwt/examples/test_client.ml
```diff
@@ -1,25 +1,50 @@
...
 let test_client _ =
+  X509_lwt.private_of_pems
+    ~cert:server_cert
+    ~priv_key:server_key >>= fun cert ->
   let port = 4433 in
   let host = "127.0.0.1" in
   X509_lwt.authenticator `No_authentication_I'M_STUPID >>= fun authenticator ->
   Tls_lwt.connect_ext
-    Tls.Config.(client ~authenticator ~ciphers:Ciphers.supported ())
+    Tls.Config.(client (* ~certificates:(`Single cert) *) ?cached_ticket:!mypsk ~ticket_cache ~authenticator ~ciphers:Ciphers.supported ())
```
cfcs: Wouldn't it be neat if the the `client` could look up the most relevant ticket in the `ticket_cache` instead of the user passing it in? (I guess the current `session_cache` is a bit geared towards a server?)

lwt/examples/resume_echo_server.ml
```diff
@@ -0,0 +1,107 @@
...
+let string_of_unix_err err f p =
+  Printf.sprintf "Unix_error (%s, %s, %s)"
+    (Unix.error_message err) f p
+
+
+module HT = Hashtbl.Make (Tls.Core.PreSharedKeyID)
+let cache_psk, psk_cache =
+  let cache = HT.create 7 in
+  ((fun ed -> HT.add cache ed.Tls.Core.psk_id ed),
+   (fun id -> if HT.mem cache id then Some (HT.find cache id) else None))
```
cfcs: As noted previously I think we should be careful about what implementations we provide for example session caches, since chances are people will copy-paste them if we don't provide a default/easy-to-use one in the library.
Thus, I believe our example session caches should be some sort of fixed-size LRU to prevent running out of memory.

lib/config.ml
hannesm: to me this sounds like a general potential improvement to the configuration validation, which should be done -- but best be done as a separate improvement to avoid this PR increasing endlessly in scope and size.

lib/config.mli
```diff
@@ -18,19 +18,33 @@ type own_cert = [
...
+  timestamp : unit -> Ptime.t
+}
+
+type ticket_cache_opt = ticket_cache option
+
 (** configuration parameters *)
 type config = private {
-  ciphers           : Ciphersuite.ciphersuite list ; (** ordered list (regarding preference) of supported cipher suites *)
+  ciphers : Ciphersuite.ciphersuite list ; (** ordered list (regarding preference) of supported cipher suites *)
+  ciphers13 : Ciphersuite.ciphersuite13 list ;
```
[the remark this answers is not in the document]
hannesm: out of scope for this PR

lib/core.ml
```diff
@@ -10,17 +10,43 @@ type tls_version =
...
+  | TLS_1_2, TLS_1_2 -> 0 | TLS_1_2, _ -> -1 | _, TLS_1_2 -> 1
+  | TLS_1_3, TLS_1_3 -> 0
+
+let next = function
+  | TLS_1_0 -> Some TLS_1_1
+  | TLS_1_1 -> Some TLS_1_2
+  | TLS_1_2 -> Some TLS_1_3
+  | TLS_1_3 -> None
+
+let all_versions (min, max) =
```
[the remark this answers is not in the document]
hannesm: I don't find your version particularly easier to understand, sorry.

lib/config.ml
hannesm: out of scope for this PR -- yes, we should separately deprecate RC4 (and maybe 3DES?)

lwt/examples/resume_echo_server.ml
hannesm: as noted above, I think this is a valid feature request, but just not in scope for this PR.

lwt/examples/test_client.ml
hannesm: maybe. maybe later.

lib/writer.ml
hannesm: when I wrote the code, I did not know whether this was the intended binary representation.

lib/packet.ml
```diff
@@ -125,22 +130,41 @@ type extension_type =
...
+  | EXTENDED_MASTER_SECRET                 [@id 23] (*RFC7627*)
+  | TOKEN_BINDING                          [@id 24] (*RFC8472*)
+  | CACHED_INFO                            [@id 25] (*RFC7924*)
+  | TLS_LTS                                [@id 26] (*draft-gutmann-tls-lts*)
+  | COMPRESSED_CERTIFICATE                 [@id 27] (*draft-ietf-tls-certificate-compression*)
+  | RECORD_SIZE_LIMIT                      [@id 28] (*RFC8449*)
+  | PWD_PROTECT                            [@id 29] (*RFC-harkins-tls-dragonfly-03*)
+  | PWD_CLEAR                              [@id 30] (*RFC-harkins-tls-dragonfly-03*)
+  | PASSWORD_SALT                          [@id 31] (*RFC-harkins-tls-dragonfly-03*)
+  | SESSION_TICKET                         [@id 35] (*RFC4507*)
```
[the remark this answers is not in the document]
hannesm: not sure what you suggest here -- do you really think it is worth to point to specific sections for all constructors? I find this a bit excessive (and additionally, in most cases the constructors have been defined in much earlier RFCs). The comments are mostly from https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml (which, in case you want to lookup/debug existing magic values, you should read / be aware of anyways).

hannesm:
I rebased on top of current master, and addressed most of the comments from @cfcs

hannesm:
now CI passes, and also the interoperability shell scripts (with openssl s_client and s_server) report success!

lib/config.mli
```diff
@@ -135,6 +159,16 @@ module Ciphers : sig
...
+
+  val psk : ciphersuite list
+  (** [psk] is a list of ciphersuites which use a pre-shared key (sublist of [default]). *)
+
+  val psk_of : ciphersuite list -> ciphersuite list
+  (** [psk_of ciphers] selects all ciphersuites which use a pre-shared key. *)
+
+  val default13 : ciphersuite13 list
+
+  val supported13 : ciphersuite13 list
```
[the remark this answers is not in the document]
hannesm: thanks, in 476b262 I removed all of that. PSK used to be in some 1.3-draft negotiated via ciphersuite, but is no longer -- now there's a hello extension for this.

lib/config.mli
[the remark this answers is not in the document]
hannesm: I changed my mind, and 476b262 addresses this.

hannesm:
force-pushed, plan to merge and release when CI is green

Issue #400, opened 2019-10-16
Need separate tls-mirage package


talex5:
At the moment, if you want to get `tls.mirage`, you have to read the opam file to find:

```
"--with-mirage" "%{mirage-flow-lwt+mirage-kv-lwt+mirage-clock+ptime:installed}%" 
```

and then add dependencies in your own opam file on `mirage-flow-lwt, mirage-kv-lwt, mirage-clock, ptime`, even though you might not be using any of them directly. This is fragile and easy to get wrong (see https://github.com/mirage/capnp-rpc/pull/181).

Ideally, tls should be split to provide a separate `tls-mirage` package with the Mirage support, so that people can just depend on that.

hannesm:
done with 0.11.0

PR #407, opened 2020-03-12
use dune, use mirage-crypto

hannesm:
splits into tls and tls-mirage package (#400 //cc @talex5)


hannesm:
tls.lwt is still part of the tls package, but lwt is no longer an optional dependency. it's a tradeoff.

PR #408, opened 2020-03-13
tls-mirage.opam: fix the name in the file

avsm:


hannesm:
thanks. I think we can as well remove the name from the opam file -- or is there a reason to keep it?

PR #417, opened 2020-07-31
Demote group/cipher logging to DEBUG

xguerin:
When using `logs` through a parent application like `opium`, cipher and group information would appear for every new connection established. When log level is set to info, 95% of the logs would be about cipher and group. It seems a good idea to restrict the scope of this information to `debug`.


hannesm:
Thanks for your PR, I'll work on merging it and prepare a tls release later in August when I'm back from vacation.

PR #419, opened 2020-09-21
Rename length to v_length to be compatible with cstruct.6.0.0

dinosaure:
Due to the introduction of `Cstruct.length` and `let open Cstruct in`, we must rename `length` to `v_length`.

### DIFF
--- a/lib/crypto.ml
+++ b/lib/crypto.ml
@@ -97,13 +97,13 @@ let adata_1_3 len =
   Cstruct.BE.set_uint16 buf 3 len ;
   buf
 
-let pseudo_header seq ty (v_major, v_minor) length =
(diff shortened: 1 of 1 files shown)


tls.opam
```diff
@@ -17,8 +17,8 @@ depends: [
   "ocaml" {>= "4.08.0"}
   "dune" {>= "1.0"}
   "ppx_sexp_conv" {>= "v0.9.0"}
-  "ppx_cstruct" {>= "3.0.0"}
-  "cstruct" {>= "4.0.0"}
+  "ppx_cstruct" {>= "6.0.0"}
+  "cstruct" {>= "6.0.0"}
```
hannesm, line 21: no need for bumping these constraints.

hannesm:
thanks

PR #421, opened 2020-11-05
Add support for OCaml 4.12

kit-ty-kate:
OCaml 4.12 added `List.compare` and now shadows `Pervasives.compare` in these particular contexts.


hannesm:
thanks, merged and released.

PR #422, opened 2020-12-02
Delete the global recv_buf and be local to each connection on tls.lwt

dinosaure:
I think currently we are not able to safely start multiple TLS connection into the same program because `recv_buf` is global. I got some data-race errors with Conduit when we try to start several clients and one server and let the scheduler (lwt or async) to re-order them eac others.

This small patch wants move locally (for each connection - client or server) the `recv_buf` used by `read_react`.

### DIFF
--- a/lwt/tls_lwt.ml
+++ b/lwt/tls_lwt.ml
@@ -39,6 +39,7 @@ module Unix = struct
                      | `Eof
                      | `Error of exn ] ;
     mutable linger : Cstruct.t option ;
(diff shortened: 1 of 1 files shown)


hannesm:
thanks, this is part of 0.12.7 (https://github.com/ocaml/opam-repository/pull/17763) with a reworded commit message 2aaf02acb792e79073b5bc0fda7d25aabe2ef3c1

Issue #413, opened 2020-06-09
Fails with UNEXPECTED_MESSAGE or HANDSHAKE_FAILURE where openssl succeeds


ryanakca:
I'm unable to connect to certain hosts using ocaml-tls. In particular, connections fail with UNEXPECTED_MESSAGE or HANDSHAKE_FAILURE. I suspect this is due to a bug in ocaml-tls for the following reasons:

1. I am able to connect to these problematic hosts with both s_client and the conduit ocaml library (using OpenSSL as the backend)
2. I am able to connect to other hosts without any issues (so I suspect it isn't user error).

Here's a minimum working example to reproduce the issue, along with the output of s_client.

```
rak@zeta:~/gemini/mwe$ cat dune
(executable
 (public_name mwe)
 (name mwe)
 (libraries base lwt lwt_ppx mirage-crypto-rng.lwt tls.lwt)
 (preprocess (pps lwt_ppx)))
rak@zeta:~/gemini/mwe$ cat mwe.ml
let connect host port =
  let authenticator ~host:_ _ = Ok None in
  let config = Tls.Config.client ~peer_name:host
      ~authenticator ~ciphers:Tls.Config.Ciphers.supported () in
  let%lwt () = Mirage_crypto_rng_lwt.initialize () in
  try%lwt
    let%lwt (ic, oc) = Tls_lwt.connect_ext config (host, port) in
    let%lwt () = Lwt_io.close ic in
    let%lwt () = Lwt_io.close oc in
    let%lwt () = Lwt_io.printl ("Success: " ^ host) in
    Lwt.return ()
  with
  | Tls_lwt.Tls_alert alert ->
    let%lwt () = Lwt_io.printl (host ^ " " ^ Tls.Packet.alert_type_to_string alert) in
    Lwt.return ()

let () =
  Lwt_main.run begin
    let hosts = [ "gemini.circumlunar.space" ; "conman.org" ] in
    let connections = List.map (fun h -> connect h 1965) hosts in
    let%lwt () = Lwt.join connections in
    Lwt.return ()
  end
rak@zeta:~/gemini/mwe$ dune build && ../_build/default/mwe/mwe.exe
Entering directory '/home/rak/gemini'
gemini.circumlunar.space UNEXPECTED_MESSAGE
conman.org HANDSHAKE_FAILURE
rak@zeta:~/gemini/mwe$ echo "" | openssl s_client -servername gemini.circumlunar.space -connect gemini.circumlunar.space:1965
[...]

PR #424, opened 2020-12-07
Improve ECC support in TLS 1.2 (RFC 8422)

hannesm:
The Go implementation of TLS dislikes client hello without a ECPointFormats
extension, so let's revert from #413 and re-add this extension with a single
value (uncompressed) to client hello (and server hello if client sent it, and
an ECC ciphersuite was selected).

In TLS 1.3 only settings, this change is a no-op.


hannesm:
merged manually

PR #428, opened 2021-03-31
Use mirage-crypto-ec for elliptic curves, support ECDSA certificates

hannesm:
This unifies the elliptic crypto primitives being used (X25519/P256), and adds support for P384 and P521 ECDH.

In addition, ECDSA (and EdDSA) certificates are now supported.


hannesm:
This is ready. I tested with the interop-sclient and interop-sserver shell scripts (using openssl on the other end).

The first commit is pretty straightforward, it uses the DH from mirage-crypto-ec for elliptic curves.
The second commit is slightly intricate, it adds ECDSA and EdDSA support. Complexity arises from interdependent configuration options and session parameters:
- Config.groups (client hello EllipticCurves / NamedCurve extension) in TLS 1.2 (and below) restricts the curves to use for DH _and_ ECDSA/EdDSA
- Config.signature_algorithms (since 1.2 [where it is optonal, required in 1.3]) specifies which signature scheme and public key type are accepted (ClientHello SignatureAlgorithms extension)
- cross-constraints for ciphersuites (1.2 and below), which specify not only the key exchange method, but also the public key type to use

This PR builds a working configuration at "Tls.Config.client/server" invocation (depending on provided values and certificates, it may restrict some parameters (ciphers/groups/signature_algorithms), may lead to an error (conflicting values provided), and during session initiation it attempts to select a working set of parameters (group, certificate, ciphersuite, signature_algorithm) -- taking the 1.0/1.1 vs 1.2 vs 1.3 differences into account.

TL;DR: The API changed slightly (some types are gone now), a server configuration may have multiple certificates for the same hostnames (but different public key types), and the interoperability with openssl has been tested.

PR #430, opened 2021-04-20
Use sexp_of ppx and remove unnecessary t_of_sexp stubs

torinnd:
[@@deriving sexp_of] seems like the more appropriate ppx to use given many of the types in lib/ can't be deserialized.

This would avoid confusion like https://github.com/mirleft/ocaml-tls/pull/398 where it's unclear if a state can be restored from an S-expression.

### DIFF
--- a/lib/reader.mli
+++ b/lib/reader.mli
@@ -6,10 +6,7 @@ type error =
   | Underflow
   | Overflow       of int
   | UnknownVersion of (int * int)
-  | UnknownContent of int
-
-val error_of_sexp : Sexplib.Sexp.t -> error
-val sexp_of_error : error -> Sexplib.Sexp.t
+  | UnknownContent of int [@@deriving sexp]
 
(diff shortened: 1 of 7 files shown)


lib/reader.mli
```diff
@@ -6,10 +6,7 @@ type error =
   | Underflow
   | Overflow       of int
   | UnknownVersion of (int * int)
-  | UnknownContent of int
-
-val error_of_sexp : Sexplib.Sexp.t -> error
-val sexp_of_error : error -> Sexplib.Sexp.t
+  | UnknownContent of int [@@deriving sexp]
```
hannesm, line 9:
```suggestion
  | UnknownContent of int [@@deriving sexp_of]
```

hannesm:
thanks, I'll get to this soon. it looks good (see a minor comment above)

torinnd:
Sure, that change makes sense. Thanks for taking a look at this!

PR #355, opened 2017-01-08
Async

vbmithr:
Do not merge yet, WIP.

I'm trying to add an Async backend to `ocaml-tls`!

PR #432, opened 2021-05-17
Async implementation, including examples

torinnd:
I'd like to offer an Async I/O implementation for Ocaml_tls. This is heavily cribbed from the existing Lwt implementation and is offered as a I/O-agnostic functor similar to cohttp.

I'm not familiar enough with Lwt to port that code over but I believe it should be relatively straightforward for someone more familiar with Lwt.

A test client and echo server are provided in an example subdirectory.

I believe this PR supersedes https://github.com/mirleft/ocaml-tls/pull/355, which doesn't appear to be actively developed.

Credit to @dinosaure  for authoring https://github.com/janestreet/ocaml-tls/pull/1 which was used for inspiration and provided some Async IO code.

### DIFF
--- a/tls-async.opam
+++ b/tls-async.opam
@@ -1,0 +1,33 @@
+opam-version: "2.0"
+homepage:     "https://github.com/mirleft/ocaml-tls"
+dev-repo:     "git+https://github.com/mirleft/ocaml-tls.git"
+bug-reports:  "https://github.com/mirleft/ocaml-tls/issues"
+doc:          "https://mirleft.github.io/ocaml-tls/doc"
+author:       ["David Kaloper <david@numm.org>" "Hannes Mehnert <hannes@mehnert.org>" "Eric Ebinger <github@eric.theebingers.com>" "Calascibetta Romain <romain.calascibetta@gmail.com>"]
+maintainer:   ["Hannes Mehnert <hannes@mehnert.org>" "David Kaloper <david@numm.org>"]
+license:      "BSD-2-Clause"
+
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "dune" {>= "1.0"}
+  "tls" {= version}
+  "x509" {>= "0.13.0"}
+  "ptime" {>= "0.8.1"}
+  "async"
+  "async_unix"
+  "core"
+  "cstruct-async"
+  "ppx_jane"
+  "mirage-crypto-rng-async"
+]
+tags: [ "org:mirage"]
+synopsis: "Transport Layer Security purely in OCaml, Async layer"
+description: """
+Tls-async provides Async-friendly tls bindings
+"""
--- a/async/dune
+++ b/async/dune
@@ -1,0 +1,5 @@
+(library
+  (name        tls_async)
+  (public_name tls-async)
+  (preprocess (pps ppx_jane))
+  (libraries async core cstruct-async mirage-crypto-rng-async tls))
--- a/async/examples/dune
+++ b/async/examples/dune
@@ -1,0 +1,11 @@
+(executable
+  (name    test_client)
+  (modules test_client)
+  (preprocess (pps ppx_jane))
+  (libraries async core tls-async))
+
+(executable
+  (name    test_server)
+  (modules test_server)
+  (preprocess (pps ppx_jane))
+  (libraries async core tls-async))
--- a/async/examples/test_client.ml
+++ b/async/examples/test_client.ml
@@ -1,0 +1,30 @@
+open! Core
+open! Async
+open Deferred.Or_error.Let_syntax
+
+let config = Tls.Config.client ~authenticator:(fun ~host:_ _ -> Ok None) ()
+
+let test_client () =
+  let host = "127.0.0.1" in
+  let port = 8443 in
+  let hnp = Host_and_port.create ~host ~port in
+  let%bind (_ : Tls_async.Session.t), rd, wr =
+    Tls_async.connect config (Tcp.Where_to_connect.of_host_and_port hnp) ~host:(Some host)
+  in
+  let req =
+    String.concat
+      ~sep:"\r\n"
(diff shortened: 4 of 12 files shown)


lib/io_intf.ml
```diff
@@ -0,0 +1,84 @@
...
+
+  val bind : 'a t -> f:('a -> 'b t) -> 'b t
+  val return : 'a -> 'a t
+  val map : 'a t -> f:('a -> 'b) -> 'b t
+
+  val tls_error : Tls_error.t -> 'a t
+end
+
+module type Fd = sig
+  module Concurrency : Concurrency
```
avsm, line 26: Isn't there already a module type in Core for bind/return/map that we could use, instead of defining a new Concurrency module signature here?  It's not quite clear to me what `tls_error` is doing in the Concurrency module type either.

torinnd: I don't think lib/ has a dependency on Core, right? I didn't want to impose opinions about which standard library to use.

dinosaure:
By principle, I don't really like the _functor_ way, specially for this case. My experience tell me that this is a bad idea to _functorize_ I/O specially in that case. More concretely, the inherent double maintenance job is needed even if we use a _functor_ due to the inherent differences between `lwt`/`async` (which are not expressed by types of signatures...).

Finally, and it's my opinion (then, I will let @hannesm to take the best solution), I prefer a _completely_ dissociate `tls.async` & `tls.lwt` where the common code will be only `tls` without any _functors_/_signatures_ which are, from my point of view, too restrictive about possibilities inherent & different between these _backends_ - and heterogeneous & complex codebases that have decided to go in this direction confirm my suspicions.

tls-async.opam
```diff
@@ -0,0 +1,33 @@
...
+build: [
+  ["dune" "subst"] {dev}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+
+depends: [
+  "ocaml" {>= "4.08.0"}
+  "dune" {>= "1.0"}
+  "tls"
```
kit-ty-kate, line 19:
```suggestion
  "tls" {= version}
```
If this is meant to be released all at the same time

lib/io_intf.ml
avsm: Right -- I think you're better off moving all the `lib/` stuff directly into `async/`. Experience has shown that functorising over Lwt/Async doesn't work well, and there are quite deep differences in how both handle concurrency (e.g. Monitors _vs_ monadic exceptions). 

Depending on your plans for the async backend, it might be better for you to keep this outside of the mirage/ org, so it can be released independently with the rest of the Core releases.  Have you considered unifying the async_ssl and this interface into one library?  If it's better to keep it in this repository, we can, but the tls library is structured such that you should be able to use the core of it externally too.

avsm:
offline discussion with @torinnd: he's going to discuss maintaining this in the janestreet/ org instead of here and update when that's figured out.

hannesm:
Dear @torinnd, thanks for your contribution. I'm fine with merging a tls-async package into this repository. IMHO it makes sense to maintain the code in this repository, so any changes to the core library will be reflected in the async layer as well.

I'm not convinced of the `lib/io.ml` and `lib/io_intf.ml`, could you fold them into the `async` subdirectory? I am aware that there is already some duplication (of rather intricate code) in `lwt/` and `mirage/` subdirectories, but I'm not convinced an `Io` functor is the way to go here.

lib/io_intf.ml
torinnd: I've moved this code out of lib/ and into async/

hannesm:
Thanks, this looks fine (note: I have not looked into detail of the async package). Would you mind to squash everything into a single commit and add a `Co-Authored-By: Calascibetta Romain <romain.calascibetta@gmail.com>` in the commit message, so @dinosaure is attributed (AFAIU the code is based on his work)?

torinnd:
Sure, I'd be happy to fix the commits once I convince travis-ci and ocaml-ci to build this. I think the error I can explain is that it's unhappy because tls.opam doesn't contain ppx_jane (or ppx_let, which I believe is the only preprocessor used in this PR). There are other things it's complaining about from the lwt/examples directory that don't make much sense to me.

Would you be open to adding ppx_let to tls.opam @hannesm? If not I can rewrite it to avoid this dependency.

hannesm:
@torinnd ppx_let is only used in async AFAICT. so eventually a `(package tls-async)` is needed for the examples (thus they're not compiled when dune build -p tls is executed).

the travis issues look like a `tls-async:.` is missing somewhere... but you can safely ignore them and focus on ocaml-ci (travis is mainly here to do some mirageos unikernel builds).

torinnd:
I think the lwt examples in mirleft:main are the only remaining issue for the ocaml-ci check. Afaict the return type of [Tls_lwt.epoch] changed from a polymorphic [Ok 'a | `Error] to a [('a, unit) result] in https://github.com/mirleft/ocaml-tls/commit/3fabf04f5c623b759d11e4990ffa0d2df9becf55. Do you want me to try and fix that in this PR as well?

hannesm:
@torinnd thanks. no need to fix this in this PR, I'll do it separately.

hannesm:
sorry, I fixed the lwt example issues in main, and then force-pushed main to your branch (where instead I intended to force-push a rebased version to your main branch). now I'm no longer allowed to push to your main branch, would you mind to rebase your commit to the current main branch and reopen this PR?

thanks.

torinnd:
It should be rebased. I don't believe I've got permission to re-open though. Maybe you need to do that @hannesm? Otherwise I can just mint another one.

hannesm:
I did a minor adjustment, `Io_intf.S.epoch` now returns a `result` (as done in #429 for Lwt and Mirage). I also added you @torinnd and @dinosaure to the authors list of tls-async.opam. When CI passes, I'll merge and release.

torinnd:
@hannesm I think the return type for Async's epoch should be an Or_error.t rather than a result to be consistent with the rest of the interface. Aside from that I'm happy with the tweaks.

hannesm:
@torinnd well spotted, I changed and force-pushed.

hannesm:
(awaiting review from @dinosaure)

dinosaure (approved):

dinosaure:
From what I see and what I did, the implementation seems fine and it follows the same design than `lwt` :+1:. It seems good to me to merge.

PR #434, opened 2021-06-22
peer_name should be strongly typed

torinnd:
Change [peer_name] to a [[\`host] Domain_name.t]. This passes the obligation to handle errors to the caller.

[peer_name] is defined as a [string option] and passed around as such until it arrives at the function [host_name_opt] in lib/handshake_common.ml, where it is converted into a [[\`host] Domain_name.t option]. A malformed string that cannot be lifted into a [[\`host] Domain_name.t] will return [None], which will disable hostname verification. This will cause, for example, connections to IP addresses like “1.2.3.4” over HTTPS to succeed where one might’ve expected them to fail.

I haven't updated the lwt/ code and expect that won't compile, so I'm tagging this as a draft. It's also an interface change and probably should be rolled into a major version bump, if the project is interested in this change.


hannesm:
the API change looks good to me -- tls predates the domain-name library, that's the reason why peer_name is a string option at the moment. I'll need to refresh my brain about length restrictions in SNI and Domain names -- and whether there are use cases where the peer name is not a host name.

lib/reader.ml
```diff
@@ -347,7 +347,9 @@ let parse_client_extension raw =
     match int_to_extension_type etype with
     | Some SERVER_NAME ->
        (match parse_hostnames buf with
-        | [name] -> `Hostname name
+        | [name] ->
+          let name = Domain_name.of_string_exn name |> Domain_name.host_exn in
```
avsm, line 351: It may be better to use one of the `raise_*` functions here in the event of a bad domain name, instead of raising an Invalid_argument (which is what Domain_name.of_string_exn will do if there's a malformed domain passed to it). That'll make it easier for users of the library to reliably catch parsing exceptions.

hannesm: @avsm good point. The exposed API (reader.mli) includes `parse_handshake` -- which does a global `catch` (defined in reader.ml - translating some exceptions into `(_, err) result`). Certainly the treatment of `Invalid_argument` could be improved, or the failing `Domain_name.of_string_exn` should be translated to an informational error message. Certainly over the last 8 years the OCaml `result` ecosystem improved a lot, eventually we can even use some Stdlib operators.

torinnd: Thanks for the feedback @avsm ! I've cleaned up the exception handling here.

torinnd:
> the API change looks good to me -- tls predates the domain-name library, that's the reason why peer_name is a string option at the moment. I'll need to refresh my brain about length restrictions in SNI and Domain names -- and whether there are use cases where the peer name is not a host name.

I think the only practical limit is enforced by [DNS](https://datatracker.ietf.org/doc/html/rfc1034#section-3.1) and limits [peer_name] to 253 bytes. The [TLS RFC](https://datatracker.ietf.org/doc/html/rfc6066#section-3) dances around the issue a bit (16 bit length for the "list" of server names, but there must only be one of each type, and DNS is the only type).

torinnd:
I think there are ocaml-ci problems, possibly with tls-async.opam? I think aside from that this PR is ready for review.

```
- cstruct-async -> (problem)
    tls-async dev requires >= v0.14
    Rejected candidates:
      cstruct-async.6.0.0: Incompatible with restriction: >= v0.14
      cstruct-async.5.2.0: Requires cstruct = 5.2.0
      cstruct-async.5.1.1: Requires cstruct = 5.1.1
      cstruct-async.5.0.0: Requires cstruct = 5.0.0
      cstruct-async.4.0.0: Requires cstruct = 4.0.0
      ...
- ocaml -> ocaml.4.12.0
    ocaml-base-compiler 4.12.0 requires = 4.12.0
- ocaml-base-compiler -> ocaml-base-compiler.4.12.0
    User requested = 4.12.0
2021-06-25 09:23.37 [INFO] = ubuntu-21.04-4.12 =
2021-06-25 09:23.37 [INFO] Can't find all required versions.
```

hannesm:
Thanks for your work, this looks good to go.

PR #435, opened 2021-07-01
Add x509-async module

torinnd:
This is a subset of the changes proposed in https://github.com/mirleft/ocaml-x509/pull/150. At a high level:
* `Cstruct.t` is replaced with `contents:string` for various decode functions
* ``('a, [ `Msg of string ])`` result is replaced with `'a Or_error.t`
* `of_pem_file` and `of_pem_dir` functions are provided where it's useful to go from a `Filename.t -> t` or `t list`
* [Authenticator] gets a [Param] submodule to facilitate constructing an `Authenticator.t` where I/O is required

This Pull Request doesn't include the sexplib or `to_string` opinions proposed in https://github.com/mirleft/ocaml-x509/pull/150 as they're relatively easy to inline.

The example server test_server.ml is updated to use the new call sites.

### DIFF
--- a/async/dune
+++ b/async/dune
@@ -2,4 +2,4 @@
   (name        tls_async)
   (public_name tls-async)
   (preprocess (pps ppx_jane))
-  (libraries async core cstruct-async mirage-crypto-rng-async tls))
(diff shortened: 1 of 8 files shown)


hannesm:
thanks.

PR #437, opened 2021-07-26
implement timeout (in effectful layers), fix #433

### LINKED ISSUE #433: add a timeout for establishing a connection
in all effectful layers, the client_of_fd / server_of_fd should fail with a timeout after a reasonable amount of time passed. otherwise there is state that is never collected when a remote client only opens a TCP connection, but never starts a tls handshake. (similarly a server may not respond on the tls layer).

hannesm:
@torinnd could you help how something along these lines would be done for the async layer?

### DIFF
--- a/lwt/tls_lwt.ml
+++ b/lwt/tls_lwt.ml
@@ -2,8 +2,7 @@ open Lwt
 
 exception Tls_alert   of Tls.Packet.alert_type
 exception Tls_failure of Tls.Engine.failure
-
-let o f g x = f (g x)
+exception Tls_timeout
 
 (* This really belongs just about anywhere else: generic unix name resolution. *)
 let resolve host service =
@@ -28,7 +27,7 @@ module Lwt_cs = struct
 
   let rec write_full fd = function
     | cs when Cstruct.length cs = 0 -> return_unit
-    | cs -> write fd cs >>= o (write_full fd) (Cstruct.shift cs)
+    | cs -> write fd cs >>= fun w -> write_full fd (Cstruct.shift cs w)
 end
 
 module Unix = struct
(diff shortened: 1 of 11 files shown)


lwt/tls_lwt.ml
```diff
@@ -141,7 +140,16 @@ module Unix = struct
           | `Eof     -> fail End_of_file
           | `Ok cs   -> push_linger t cs ; drain_handshake t
 
-  let reneg ?authenticator ?acceptable_cas ?cert ?(drop = true) t =
+  let default_timeout = Duration.of_sec 60
+
+  let with_timeout timeout f =
+    if timeout > 0L then
```
torinnd, line 146: I'm a bit sad about making a time span of zero a magic value that disables timeouts. Other time libraries I've worked with generally provide something like [Time.max_value] if you really want to never time out.

I don't think I'd want to implement a similar check in tls_async, and would weakly prefer it went away here to keep the behavior relatively consistent across concurrency abstractions.

hannesm: ok, we can as well use `Int64.max_int` and no conditional in `with_timeout`. Thanks for the suggestion.

torinnd:
> @torinnd could you help how something along these lines would be done for the async layer?

I'd be happy to! I've got something compiling that roughly approximates the Lwt changes. I'll try to clean it up and get it into a presentable state. I don't believe I am allowed to contribute to this PR, however. Should I offer a PR to https://github.com/hannesm/ocaml-tls/tree/timeout ?

lwt/tls_lwt.ml
hannesm: done in 507fade

hannesm:
> Should I offer a PR to https://github.com/hannesm/ocaml-tls/tree/timeout ?

That would be great, thanks! :) Once this is done, I'm fine to cut a release.

torinnd:
https://github.com/hannesm/ocaml-tls/pull/4 is ready for review

hannesm:
the reason for this PR was due to potential issues (in memory consumptions) of some MirageOS unikernels (e.g. https://github.com/roburio/tlstunnel) - which after testing is not the case. I'll hold back this PR for now, and investigate the memory profile of the unikernels in question.

hannesm:
Closing, this may be done differently on a different day.

PR #442, opened 2021-11-05
Drop dependency on Async_find

torinnd:
I propose we remove the single invocation of [Async_find] in [X509_async] to drop the dependency.

This also has the nice property of truing up the Async version with the existing Lwt version by dropping support for nested directory structures.

### DIFF
--- a/async/dune
+++ b/async/dune
@@ -2,4 +2,4 @@
   (name        tls_async)
   (public_name tls-async)
   (preprocess (pps ppx_jane))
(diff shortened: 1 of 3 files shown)


hannesm:
looks fine to me, would you like a release with this patch merged?

torinnd:
If it's convenient!

hannesm:
done in 0.15.2 / https://github.com/ocaml/opam-repository/pull/20011 / already merged into opam-repository.

PR #443, opened 2021-12-30
Add an unix implementation of TLS using `Unix.file_descr`

dinosaure:
This PR is a draft because I missed `key_update` and `reneg`. Currently, the implementation follows `tls-mirage` as much as possible and the diff between them is smaller as I can. The resulted interface follows `Mirage_flow.S` without `'a Lwt.t` to smoothly use it into some others MirageOS projects (such as `colombe` or `dns`).

- [x] `key_update`
- [x] `reneg`
- [ ] `client_of_flow` should do the `connect`

### DIFF
--- a/unix/dune
+++ b/unix/dune
@@ -1,0 +1,5 @@
+(library
+ (name tls_unix)
+ (public_name tls.unix)
+ (wrapped false)
+ (libraries tls unix ptime.clock.os mirage-crypto-rng.unix))
--- a/unix/tls_unix.ml
+++ b/unix/tls_unix.ml
@@ -144,6 +144,35 @@ let read socket =
             drain_handshake flow
         | `Error e -> Error e
         | `Eof     -> Error Closed
+
+  let reneg ?authenticator ?acceptable_cas ?cert ?(drop = true) flow =
+    match flow.state with
+    | `Eof        -> Error Closed
+    | `Error e    -> Error e
+    | `Active tls ->
+        match Tls.Engine.reneg ?authenticator ?acceptable_cas ?cert tls with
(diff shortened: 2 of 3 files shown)


dinosaure:
An other possibility is to abstract the underlying flow such as:
```ocaml
type 'flow Tls_unix.flow

type 'flow fully_write = 'flow -> string -> int -> int -> unit
type 'flow read = 'flow -> [ `Data of string | `Eof ]

val client_of_flow : fully_write:'flow fully_write -> read:'flow read -> Tls.Config.client -> ?host:[ `host ] Domain_name.t ->
  'flow -> ('flow flow, error) result
```

hannesm:
Nearly 3 years later, what should we do with this? I know there (is/was) a similar module in https://github.com/cfcs/ofetch/

reynir:
With #503 merge I guess this can be closed?

dinosaure:
> With https://github.com/mirleft/ocaml-tls/pull/503 merge I guess this can be closed?

It's more to do with the Unix module only. I don't have a problem with keeping this PR open: it's just that I've found myself several times having to make this module for different cases in which I didn't want to use a scheduler. Maybe the best thing (if you're not a maniac about open PRs) is to leave this PR open as an example.

reynir:
Ah ok. I was just going off the comment in #503:

> https://github.com/mirleft/ocaml-tls/pull/443 rebased with `string`.

hannesm:
> Ah ok. I was just going off the comment in #503:
> 
> > #443 rebased with `string`.

Thanks for spotting the typo, it was supposed to read #494 instead of #443 I assume.

PR #448, opened 2022-07-30
Expose async tls_handler

mbacarella:
I proprose this PR to expose the async tls_handler to users of the library. This is needed if you want to do your own Tcp service management, such as if you're plugging async tls support into other frameworks.


hannesm:
that's a pretty huge diff for the description. any chance you could remove any code formatting changes from this PR?

also, maybe @torinnd could take a look at the changes?

torinnd:
I'd probably recommend collapsing the new test_server_low_level example into the test_server. They don't look different enough to me to justify persisting both.

I also think I'd refactor the MLI as follows:

```
type 'a io_handler = Reader.t -> Writer.t -> 'a Deferred.t
type 'a tls_handler = Session.t -> 'a io_handler

val upgrade_server_handler
  : config:Tls.Config.server
  -> 'a tls_handler
  -> 'a io_handler
  ```

Notably this excludes any mention of the socket address from the type signature as that isn't relevant to the TLS upgrade being performed here. Callers would probably do something like `upgrade_server_handler ~config (my_cool_handler socket)` to satisfy the `Async.Tcp.Server.create` interface.

**Comment by mbacarella** [comment]
Thanks for the review. Collapsing the test servers into one and the .mli refactor done.

I believe all of the extraneous code formatting has been removed (mostly in the dune file).

File: async/tls_async.mli [hunk 1]
```diff
@@ -9,6 +9,23 @@ module Session = Session
     certificates, such as loading from a Unix filesystem *)
 module X509_async = X509_async
 
+(** [upgrade_server_handler] is what [listen] calls to handle each client.
+    It is exposed so that low-level end-users of the library can use tls-async
+    inside of code that manages Tcp services directly.
```
hannesm, line 14: that documentation string is now attached to the wrong value, and since it refers to listen, maybe push it after the listen value?

mbacarella: ah, good catch. moved it below listen as well.

hannesm:
Thanks for your contribution. I squashed and merged, this will be part of the next release.

Issue #452, opened 2022-09-27
close and shutdown semantics


hannesm:
As discussed in #451:

- TLS 1.0: only properly closed connections could be used for resumption
- TLS 1.0 - 1.2: a peer receiving a close_notify immediately has to drop everything and reply with a close_notify
- TLS 1.3: close_notify is single direction only.

The implementation does not track `close_notify` being sent or received, but it replies to a close_notify (and any other alert) with a `close_notify` itself. To support `shutdown Write` (and return `` `Eof `` early if the other peer already sent `close_notify`) we need to track this in the engine state.

bikallem:
@hannesm @talex5 
>The implementation does not track close_notify being sent or received, but it replies to a close_notify (and any other alert) with a close_notify itself. To support shutdown Write (and return `Eof early if the other peer already sent close_notify) we need to track this in the engine state.

As I understand it, in TLS 1.3, `close_notify` is not really an `EOF` - the end of the file or full shutdown. It only means that the party sending `close_notify` will only close the write side of the connection. It should still be able to read from the same connection. Additionally, we no longer need to send `close_notify` in response to a receipt of `close_notify` nor should we expect it. Here is the paragraph from the spec which led me to come to this understanding:

```
Each party MUST send a "close_notify" alert before closing its write
side of the connection, unless it has already sent some error alert.
This does not have any effect on its read side of the connection.
```

1) It would be preferable for the `Tls.Engine.state` to track receipt/send of 'close_notify` so as to reduce/eliminate duplication of logic from effectful implementations such as eio, lwt, async, unix etc. In my PR https://github.com/mirleft/ocaml-tls/pull/458, I attempt to implement such a tracking. I imagine other effectul implementations will have to track similarly if they want to implement precise 'close_notify' semantics. Therefore, `Tls.Engine.state/handle_tls` could ease effectful implementations by incorporating this logic in the Tls.Engine.state itself.

[...]

PR #456, opened 2022-11-09
tls-eio: add fuzz tests using crowbar

talex5:
The test picks two random strings, one for the client to send and one for the server. It then runs a send and receive fiber for each end.

A dispatcher fiber then sends commands to these worker fibers. It may:

- Ask a send fiber to send some plaintext from the message.
- Ask a receive fiber to read some plaintext.
- Ask the network to transmit some number of bytes in some direction.
- Ask a send fiber to shut down the connection.
- Yield (allowing the workers to make progress).

At the end, it asks any remaining send fibers to shut down, allows all network traffic to be transmitted, and waits for the receive fibers to end. It then checks that the receiver got everything that was sent (with a special case to avoid reporting errors due to #452).

By default, this just runs a few random examples. To use afl-fuzz:
```sh
opam switch create 5.0.0~beta1+afl ocaml-variants.5.0.0~beta1+options ocaml-option-afl

dune runtest   # (build the tests and do a quick run first)
mkdir input
echo hi > input/foo
cp certificates/server.{key,pem} .
afl-fuzz -m 1000 -i input -o output ./_build/default/eio/tests/fuzz.exe @@
```
If you revert #454, it finds the bug (even without afl-fuzz). It could be extended in future to test other features, such as `reneg` and `key_update`.

### DIFF
--- a/eio/tests/dune
+++ b/eio/tests/dune
@@ -11,3 +11,19 @@
     server-ec.key
     (package tls-eio)
     (package eio_main)))
+
+; "dune runtest" just does a quick run with random inputs.
+;
+; To run with afl-fuzz instead:
+;
+; dune runtest
+; mkdir input
+; echo hi > input/foo
+; cp certificates/server.{key,pem} .
+; afl-fuzz -m 1000 -i input -o output ./_build/default/eio/tests/fuzz.exe @@
+(test
+  (package tls-eio)
+  (libraries crowbar tls-eio eio.mock logs logs.fmt)
+  (deps server.pem server.key)
+  (name fuzz)
+  (action (run %{test} --repeat 200)))
--- a/eio/tests/fuzz.ml
+++ b/eio/tests/fuzz.ml
@@ -1,0 +1,271 @@
+open Eio.Std
+
+let src = Logs.Src.create "fuzz" ~doc:"Fuzz tests"
+module Log = (val Logs.src_log src : Logs.LOG)
+
+module W = Eio.Buf_write
+
+type transmit_amount = Mock_socket.transmit_amount
+
+type op =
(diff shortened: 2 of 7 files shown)


hannesm:
Thanks for your contribution. Since I'm pretty slow, would you mind to express what kind of failures you're after?

It looks you don't fuzz the input (for Reader), neither the TLS state machine (when which message is expected), but mostly the exposed (in tls-eio) state machine (i.e. once an error (or close) occured, there won't be any further data transported / all function fail). Is this correct?

Or rephrased, do you have an intuition (or actual measurement) about the coverage of this approach?

talex5:
I added `(instrumentation (backend bisect_ppx))` to the dune file and ran:
```
dune runtest --instrument-with bisect_ppx --force ./eio
bisect-ppx-report html
```

That reports 46% coverage for `tls_eio.ml` without the fuzz tests.

Annoyingly, you don't get coverage for fuzz tests because both bisect-ppx and crowbar use `at_exit` handlers, and they run in the wrong order (dumping the coverage before running the tests). But with a patched crowbar, it gets 54%. Seems to be mostly error paths missed (as expected) though some of the `linger` cases aren't hit.

hannesm:
Great, thanks for the numbers and clarification.

hannesm:
@talex5 we can merge at any time, I'd appreciate to preserve some of the discussion in this PR as a comment in the test (since once the PR is merged, it's difficult to find it again) -- i.e. what the purpose is and what is fuzz-tested here (including some preliminary coverage stats).

talex5:
> I'd appreciate to preserve some of the discussion in this PR as a comment in the test

I've added a comment to the top of the file now.

hannesm:
Thanks @talex5, merged.

Issue #457, opened 2022-12-01
EIO: Getting `Tls Failure (Fatal MACMismatch)


bikallem:
Somehow I am getting intermittent tls errors such as `TLS failure: (Fatal MACMismatch)` or `TLS failure: (Fatal (RecordOverflow 26678))` when using `tls-eio` in my repo: https://github.com/bikallem/ocaml-dns/tree/eio-tls 

This happens after I successfully establish tls connection - I can observe that I can do a successful tls handshake and can send application data over the tls connection. The first couple of data packets are sent and received okay however, it seems anything more than that seems to fail with the errors enumerated above. Could you please advise if I am using the tls-eio correctly?

bikallem:
I made another branch which uses `eio-ssl` instead of `tls-eio` and I am unable to reproduce the error I get with `tls-eio`. https://github.com/bikallem/ocaml-dns/tree/eio-ssl 

/cc @talex5

talex5:
Can't see anything obviously wrong (what are you using the `ticket_cache` for though?).

I suggest:

1. Use tcpdump or similar to get a pcap dump of a case where it goes wrong.
2. Make a test case that just replays that dump.
3. Keep simplifying the test-case as long as it keeps failing.
4. Turn on debug logging for TLS.

Maybe you're doing overlapping reads or overlapping writes somewhere? You could put a mutex around those operations to check.

Finally, does it work with the lwt+tls version of Dns?

bikallem:
> Can't see anything obviously wrong (what are you using the ticket_cache for though?)

It is the same with or without.

hannesm:
Dear @bikallem, thanks for the report. Could you in a bit more detail elaborate to which host (IP/hostname and port number) you're trying to connect?

Does a connection to the same host using tls-lwt or tls-mirage lead to the same issues? Or is it only the tls-eio package that shows these issues? If the latter, would you mind to elaborate which versions of ocaml, eio, and tls you're using?

talex5:
I had a go at reproducing this using your `eio-tls` branch. It sometimes works and sometimes crashes with an end-of-file error for me. That seems to be in some custom parsing code that no longer exists (or has changed a lot) on your main `eio` branch.

Does this still happen with the current code? It would be good to have step-by-step instructions to see the error with the latest version if so. Perhaps a Dockerfile that builds and runs the test?

[...]

PR #458, opened 2022-12-12
Dns client eio

bikallem:
Fixes https://github.com/mirleft/ocaml-tls/issues/457

Summary of Changes:

1) Simplified `t.state` to just be `Tls.Engine.state`. Previously handling and re-raising exceptions was giving obfuscating errors in tls-eio, such as  `Cancelled: Eio__core__Fiber.Not_first` or `mutex poisioned` and so forth. This and using them together with `Eio.Fiber.any/first` seemingly give wrong, unhelpful and unproductive errors. Therefore, the code deviates from `tls_lwt` style of state management. I believe the new state handling is much more direct and straightforward as preferred by eio and OCaml effects. 

2) I was also getting weird intermittent errors such as `MACMismatch` or `RecordOverflow`. From investigation, I concluded that the errors were due to the interleaving of read/write operations of the concurrent fibers, i.e. a read on tls flow means read/write on the underlying flow `t.flow`. Therefore the underlying read/write on `t.flow` should be protected with a mutex such that the read/write on `t.flow` is atomic.

3) I have also updated some deprecated function usage such as `Eio.Flow.read` to `Eio.Flow.single_read`.

/cc @talex5 @hannesm

### DIFF
--- a/eio/tls_eio.ml
+++ b/eio/tls_eio.ml
@@ -6,11 +6,15 @@ exception Tls_failure of Tls.Engine.failure
 module Raw = struct
 
   type t = {
-    flow           : Flow.two_way ;
-    mutable tls    : Tls.Engine.state ;
-    mutable linger : Cstruct.t ;
-    recv_buf       : Cstruct.t ;
-    mutex          : Eio.Mutex.t;
+    flow                      : Flow.two_way ;
+    recv_buf                  : Cstruct.t ;
+    mutex                     : Eio.Mutex.t;
+    mutable tls               : Tls.Engine.state ;
+    mutable linger            : Cstruct.t ;
+
+    (* We track 'close_notify' here since Tls.Engine.state doesn't track it. 
+       https://github.com/mirleft/ocaml-tls/issues/452 *)
+    mutable close_notify_sent : bool ;
   }
 
   let write_t t cs = Flow.copy (Flow.cstruct_source [cs]) t.flow
@@ -29,16 +33,24 @@ module Raw = struct
         begin match state with
           | `Ok tls ->
             t.tls <- tls ;
-            Option.iter (write_t t) resp ;
-            Eio.Mutex.unlock t.mutex ;
-            application_data
+            Option.iter (write_t t) resp
           | `Eof ->
             (* received "close_notify" alert from peer so shutdown receving data
-              from the peer socket. https://www.rfc-editor.org/rfc/rfc8446#section-6.1 *)
+               from the peer socket. https://www.rfc-editor.org/rfc/rfc8446#section-6.1 *)
             Eio.Flow.shutdown t.flow `Receive ;
-            None
+
+            (* We set it to true since 'Tls.Engine.handle_tls t.tls data' returns 'close_notify'
+               as `Response resp when it encounters `close_notify` in data. *)
+            if not t.close_notify_sent then begin
+              Option.iter (write_t t) resp ;
+              t.close_notify_sent <- true
+            end
           | `Alert a -> raise (Tls_alert a)
-        end
+        end ;
+
+        Eio.Mutex.unlock t.mutex ;
+        application_data
+
       | Error (failure, `Response resp) ->
         write_t t resp ;
         raise (Tls_failure failure)
@@ -100,9 +112,12 @@ module Raw = struct
       write_t t buf
 
   let close_tls t =
-    let (tls, buf) = Tls.Engine.send_close_notify t.tls in
-    write_t t buf ;
-    t.tls <- tls
+    if not t.close_notify_sent then begin
+      let (tls, buf) = Tls.Engine.send_close_notify t.tls in
+      write_t t buf ;
+      t.tls <- tls ;
+      t.close_notify_sent <- true ;
+    end
 
   let shutdown t = function
     | `Send -> close_tls t ; Flow.shutdown t.flow `Send
@@ -115,7 +130,8 @@ module Raw = struct
         flow     = (flow :> Flow.two_way) ;
         linger   = Cstruct.empty ;
         recv_buf = Cstruct.create 4096 ;
-        mutex    = Eio.Mutex.create ()
+        mutex    = Eio.Mutex.create () ;
+        close_notify_sent = false ;
       }
 
   let client_of_flow config ?host flow =
@@ -129,7 +145,8 @@ module Raw = struct
       flow     = (flow :> Flow.two_way) ;
       linger   = Cstruct.empty ;
       recv_buf = Cstruct.create 4096 ;
-      mutex    = Eio.Mutex.create ()
+      mutex    = Eio.Mutex.create () ;
+      close_notify_sent = false ;
     }
     in
     write_t t init ;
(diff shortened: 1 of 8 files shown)


eio/tls_eio.ml
```diff
@@ -5,198 +5,143 @@ exception Tls_failure of Tls.Engine.failure
...
+            Eio.Mutex.unlock t.mutex ;
+            application_data
+          | `Eof -> raise End_of_file ;
+          | `Alert a -> raise (Tls_alert a)
+        end
+      | Error (failure, `Response resp) ->
+        write_t t resp ;
+        raise (Tls_failure failure)
+    with 
+    | ( End_of_file | Tls_alert _ | Tls_failure _ ) as ex ->
```
bikallem, line 42: I am not sure but it seems when handling exceptions in effects/eio, we need to be more explicit as to what we are handling. At one time exceptions such as  this `https://github.com/ocaml/ocaml/blob/trunk/stdlib/effect.ml#L59` was being trapped in here. Perhaps this should go in the `eio` error handling guidance ?

eio/tls_eio.ml
```diff
@@ -5,198 +5,143 @@ exception Tls_failure of Tls.Engine.failure
...
-    | `Eof        -> raise End_of_file
-    | `Active tls ->
-        match Tls.Engine.send_application_data tls css with
-        | Some (tls, tlsdata) ->
-            ( t.state <- `Active tls ; write_t t tlsdata )
-        | None -> invalid_arg "tls: write: socket not ready"
-
-  let write t cs = writev t [cs]
-
-  (*
```
bikallem, line 102: This is no longer applicable since the only place we need to worry about interleaving read/write is when reading from the tls_eio flow.

eio/tls_eio.ml
```diff
@@ -5,198 +5,143 @@ exception Tls_failure of Tls.Engine.failure
...
-  let rec read t buf =
-
-    let writeout res =
-      let open Cstruct in
-      let rlen = length res in
-      let n    = min (length buf) rlen in
-      blit res 0 buf 0 n ;
+  let rec read t flow_buf : int =
+
+    let write_application_data data =
```
hannesm, line 48: from the name of this function, this sounds like "only application data will ever be written here", but AFAICT there could as well be handshake data that is being written. how did you choose the name?

eio/tls_eio.ml
```diff
@@ -5,198 +5,143 @@ exception Tls_failure of Tls.Engine.failure
...
+    write_t t buf ;
+    t.tls <- tls
+
   let shutdown t = function
-    | `Send -> close_tls t
-    | `All -> close_tls t; Flow.shutdown t.flow `All
-    | `Receive -> ()  (* Not obvious how to do this with TLS, so ignore for now. *)
+    | `Send -> close_tls t ; Flow.shutdown t.flow `Send
+    | `All -> close_tls t ; Flow.shutdown t.flow `All
+    | `Receive -> close_tls t ; Flow.shutdown t.flow `Receive
```
bikallem, line 107: Not sure about this one but I have added a corresponding close on the `t.flow` believing this version to be more semantically correct than the original. However, please advise if this is not the case.

eio/tls_eio.ml
```diff
@@ -5,198 +5,143 @@ exception Tls_failure of Tls.Engine.failure
...
-    in
-    match t.state with
-    | `Active tls when not (Tls.Engine.handshake_in_progress tls) ->
-        t
-    | _ ->
-        let cs = read_react t in
-        push_linger t cs; drain_handshake t
+    if not (Tls.Engine.handshake_in_progress t.tls) then t
+    else
+      let application_data = read_t t in
```
hannesm, line 75: Again, I don't think the naming is correct here. the buffer returned by `read_t` is some encrypted TLS data, no? and this may be handshake, application data, alerts, heartbeat, ..

hannesm: oops, sorry, `read_t` now indeed returns application data.

talex5:
I think it would be good to merge the fuzz tests (https://github.com/mirleft/ocaml-tls/pull/456) before making any further changes to tls-eio. It would also be good to modify the fuzz tests (in this PR) to show the problem being fixed (e.g. by having the fuzz tester randomly cancel reads).

> Previously handling and re-raising exceptions was giving obfuscating errors in tls-eio, such as Cancelled: Eio__core__Fiber.Not_first or mutex poisioned and so forth. 

`Cancelled` can probably be ignored (I was planning to test that once the fuzz tests were merged), but I think breaking the Tls flow on other errors is reasonable.

> From investigation, I concluded that the errors were due to the interleaving of read/write operations of the concurrent fibers, i.e. a read on tls flow means read/write on the underlying flow t.flow.

That's interesting. Since the fuzz tests don't detect this, I guess it works as long as the underlying flow orders writes correctly, which perhaps the real flow does not. Possibly we should fix it there instead.

hannesm:
What @talex5 says, but even more convincing would be a regression test (rather than a fuzz test) of the bad record mac behaviour -- if I understand both of you correctly, this should be easy since with eio everything can be mocked (including the scheduler)

bikallem:
> What @talex5 says, but even more convincing would be a regression test (rather than a fuzz test) of the bad record mac behaviour -- if I understand both of you correctly, this should be easy since with eio everything can be mocked (including the scheduler)

Yes, a test case would indeed be nice. However, I am not too familiar with crowbar or property based testing in general; so it will probably be a while before I get to it. Alternately, I was thinking about using `qcheck-lin` - https://ocaml-multicore.github.io/multicoretests/0.1/qcheck-lin/. If I remember correctly @jmid mentioned (in the 2022 tarides retreat) that the tool has some advantages with regards to testing effects(eio) based concurrency algorithms.

As an aside, do either of you know if tla+ (https://learntla.com/) is a good candidate for uncovering concurrency issues such as what this PR is trying to address?

hannesm:
Dear @bikallem, thanks again for this PR. It would really benefit if you could focus on a single issue (and best option to provide a test case and a minimal change that fixes it), and not have lots of other changes in the same PR - esp. the ``close_notify`` tracking is nothing that I'm keen to have in eio only, as mentioned in #452 this is also TLS protocol version dependent (and should properly be done in the core tls library, not in one of the effectful layers).

I'd appreciate to first see some discussion about "why a mutex is needed" and what the desired parallelism story of eio is.

hannesm:
Also, please feel free to reach out via eMail if you'd like to schedule a more instant chat (i.e. a video meeting), where we can discuss your goals and the road towards them. :)

bikallem:
> Dear @bikallem, thanks again for this PR. It would really benefit if you could focus on a single issue (and best option to provide a test case and a minimal change that fixes it), and not have lots of other changes in the same PR - esp. the close_notify tracking is nothing that I'm keen to have in eio only, as mentioned in https://github.com/mirleft/ocaml-tls/issues/452 this is also TLS protocol version dependent (and should properly be done in the core tls library, not in one of the effectful layers).

Thanks for the feedback. Indeed, this PR was supposed to only fix issue https://github.com/mirleft/ocaml-tls/issues/457. However, while attempting to write a test for the issue, I noticed another issue with my PR of not handling `close_notify` TLS message correctly. This had to be fixed in order to pass the CI. However, I agree `close_notify` is better handled by the engine. I will comment my thinking on the issue there rather than here to reduce noise. I am still working on the reproduction of the issue I faced with `ocaml-dns-client`. I will ping you once I am able to do it. 

I have converted the PR into draft for now. 

Many thanks to @talex5 for the fuzz test. I feel I am much more productive with such test harness.

bikallem:
Closing this PR for now in support of https://github.com/mirleft/ocaml-tls/issues/464

PR #459, opened 2022-12-15
tls-eio: allow cancelling reads

talex5:
Previously, Tls would treat any exception from a read as a permanent error, causing all future operations to fail. However, it is reasonable to cancel a read and then try again later. I extended the fuzz tests to check for this case.

Note that the Lwt version appears to have the same problem. Possibly we should only be catching `Eio.Io` exceptions in the first place, or maybe none at all.

The first commit cleans up the fuzz testing a bit (see the commit comment for details) and the second one adds support for cancellation.

One place I'm not sure about is when `drain_handshake` is called from `reneg`. Is it OK to abort there? I don't know what the function does and it's already marked with `XXX`. In the normal case (when `drain_handshake` is called during flow creation) it's safe because the caller never gets the `Tls_eio.t`.

I'm not sure whether we really want to encourage users to cancel reads and resume later, but this should at least fix the confusing `Cancelled: Eio__core__Fiber.Not_first` errors reported by @bikallem in #458.

### DIFF
--- a/eio/tls_eio.ml
+++ b/eio/tls_eio.ml
@@ -29,7 +29,7 @@ module Raw = struct
       raise exn
 
   let write_t t cs =
-    try Flow.copy (Flow.cstruct_source [cs]) t.flow
+    try Flow.write t.flow [cs]
     with exn ->
       (match t.state with
        | `Error _ | `Eof -> ()
--- a/eio/tls_eio.ml
+++ b/eio/tls_eio.ml
@@ -22,6 +22,9 @@ module Raw = struct
     | End_of_file as ex ->
       t.state <- `Eof;
       raise ex
+    | Eio.Cancel.Cancelled _ as ex ->
+      (* No need to break the flow just because a read got cancelled. *)
+      raise ex
     | exn ->
       (match t.state with
        | `Error _ | `Eof -> ()
--- a/eio/tests/dune
+++ b/eio/tests/dune
@@ -14,7 +14,7 @@
 
 ; "dune runtest" just does a quick run with random inputs.
 ;
-; To run with afl-fuzz instead:
+; To run with afl-fuzz instead (make sure you have a compiler with the afl option on!):
 ;
 ; dune runtest
 ; mkdir input
--- a/eio/tests/fuzz.ml
+++ b/eio/tests/fuzz.ml
@@ -39,6 +39,7 @@ type op =
   | Send of int                   (* The application sends some bytes to Tls *)
   | Transmit of transmit_amount   (* The network sends some types to the peer *)
   | Recv                          (* The application tries to read some data *)
+  | Cancel_recv                   (* The application aborts any current read *)
   | Shutdown_send                 (* The application shuts down the sending side *)
 
 let label name gen =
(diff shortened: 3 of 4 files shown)


talex5:
afl-fuzz finds that this can fail, because a TLS read operation may need to perform a write. So when the application cancels a TLS read, it may actually end up cancelled a write, which we can't recover from.

eio/tls_eio.ml
```diff
@@ -22,14 +22,17 @@ module Raw = struct
     | End_of_file as ex ->
       t.state <- `Eof;
       raise ex
+    | Eio.Cancel.Cancelled _ as ex ->
+      (* No need to break the flow just because a read got cancelled. *)
+      raise ex
     | exn ->
```
bikallem, line 28: Wouldn't it make sense to completely remove this bit instead? I believe this is trapping other exceptions in addition to `Cancelled`, such as 'Poisioned' and intermittently this https://github.com/ocaml/ocaml/blob/trunk/stdlib/effect.ml#L59

eio/tls_eio.ml
```diff
@@ -22,14 +22,17 @@ module Raw = struct
...
       (match t.state with
        | `Error _ | `Eof -> ()
        | `Active _ -> t.state <- `Error exn) ;
       raise exn
```
bikallem, line 32: Wouldn't it make sense to completely remove this bit instead? I believe this is trapping other exceptions in addition to `Cancelled`, such as 'Poisioned' and intermittently this https://github.com/ocaml/ocaml/blob/trunk/stdlib/effect.ml#L59

talex5: I assume the idea is that the flow should stop working on any unexpected error. Otherwise, you might fail to read a packet, retry, and then hang (because the sender thinks you already have the data). Possibly we should call `Printexc.get_raw_backtrace` here and store that in `t.state` too. Then you'd know where the original error came from.

But note that the original exception is still raised, with the original backtrace, the first time. So the first error you get should include the correct location in all cases. It's just if you keep trying to use it after an error then future errors don't give the exact location.

bikallem: >But note that the original exception is still raised, with the original backtrace, the first time. So the first error you get should include the correct location in all cases. It's just if you keep trying to use it after an error then future errors don't give the exact location.

Indeed, that's what I expected. However, I was getting other exceptions being reported - as mentioned in the comment above - as being the cause of error. As a user of `tls-eio`, that led me to unproductive debugging sessions since the actual error was quite something else. I have a feeling that the interaction of `Eio.Fiber.any/first`, tls_eio exception handling here  and the like is giving unpredictable exceptions. I was only able to move forward with identifying the correct issue after I removed this catch all exception handling. Removing the catch all exception handling also don't seem to impact any observable tls-eio behaviours, i.e. the tests pass and End_of_file exception is correctly reported to users of tls-eio.

talex5:
Closing, as we decided not to do this (https://github.com/mirleft/ocaml-tls/issues/464#issuecomment-1416776080):
> I agree, we can stick (at least for now) to "cancelling an operation on a flow cancels that flow". Again, I don't see a concrete use case to allow cancelling a read/write and afterwards using the flow for a different (or the same) purpose.

PR #461, opened 2022-12-16
tracing: expose log src

bikallem:
Expose log source so that we can control log level individually. This should help making debugging sesssion a bit more manageable and productive, i.e. sometimes when debugging we want to look at logs of a particular dependency a bit more deeply and at other times an `info` level suffices. 

I am finding this to be a bit helpful while debugging https://github.com/mirleft/ocaml-tls/issues/457 where trying to understand and pinpoint error among dns, tls and eio seems very noisy and unproductive with global debug level.


hannesm:
thanks

PR #462, opened 2022-12-19
misc: develop against vendored dependencies

bikallem:
This PR allows the dev repo to support developing, testing and debugging against vendored dune dependencies. 

/cc @hannesm


dune
```diff
@@ -0,0 +1 @@
+(vendored_dirs vendor)
```
avsm, line 1: This isn't necessary -- when I develop in a vendored tree, I just put this stanza inside a dune file in the vendor directory. See for example https://github.com/avsm/eeww/blob/main/lib/dune.

bikallem: Very nice, thanks! This is done now.

.gitignore
```diff
@@ -14,4 +14,5 @@ mirage/*/tls_server
 Makefile
 key_gen.ml
 
-random/
\ No newline at end of file
+random/
+vendor/
```
avsm, line 18: You don't need this any more either...

bikallem: hmmm ... git seems to want to include things in this folder and it always lists it when doing `git status`. Of course you can do selective `git add ...` but mostly I prefer the convenience of `git add --all`. This prevents from mistakenly adding things from `vendor` dir  to this repo.

hannesm:
Since there's quite some discussion about "vendoring" and "dependencies", is that somewhere documented what needs to be done and how to address issues? Is there a general move in the OCaml and opam ecossytem to add at random places "vendor"? I'm not a huge fan, especially since I've a hard time understanding what this "vendoring" is trying to solve, and how it works workflow-wise (including upstreaming diffs, etc.).

eio/tests/dune
```diff
@@ -10,6 +10,7 @@
     server-ec.pem
     server-ec.key
     (package tls-eio)
+    (package mirage-crypto-rng-eio)
```
hannesm, line 13: this change looks like an actual independent bugfix - which is fine to be merged in a separate commit (and PR)

hannesm:
Likely there exist some "vendoring" documentation, so maybe @bikallem or @avsm would mind to point to that? Is this "opam-monorepo" a part of it (or "opam lock")? I'm still puzzled, but maybe I just settled on a very arcane development style. Would love to hear how you people are working and which tools you're using.

eio/tests/dune
bikallem: I think this is only necessary in a vendoring situation like this PR, thus the change. Outside of the use-case, i.e. vendoring your dependencies - this is not needed as before this PR.

bikallem:
The vendoring approach used here is documented here https://dune.readthedocs.io/en/stable/dune-files.html#vendored-dirs-since-1-11. 

I am using vendoring here mainly so that I can insert 'printf traces/Eio.traceln' to tls-eio dependencies  located in 'vendor' dir. This PR makes working with such vendored dependencies a little bit precise.

hannesm:
Since @talex5 contributed the "eio/tests/dune", and my knowledge of mdx and dune is rather limited, I'd like to hear his opinion whether this PR is reasonable. Thanks.

talex5 (approved):
Looks good to me. When used without vendoring, the opam metadata will ensure that `mirage-crypto-rng-eio` is available, but if dune is building that at the same time then it will need to know about the dependency.

PR #463, opened 2022-12-21
tls-eio: improve fuzz tests

talex5:
These are the improvements to the fuzz tests that were part of #459. I think it's worth merging these on their own, since that PR isn't ready.

- The mock_socket has changed slightly. Previously, it would request a size to read, do the read, and then wait for another size for the next read, which would typically just be the EOF for that request. This made it harder for the fuzzer to reach useful states. Now you can do a single transmit with a single test action.

- ~Simplified the write in Tls_eio slightly by using the new `Flow.write` operation.~

- Only log a send when we actually send some data.

- Allow the fuzzer to perform all the actions needed for the Tls handshake in one go instead of requiring it to find them by searching. This provides it a way to get to the interesting cases quickly.

### DIFF
--- a/eio/tests/fuzz.ml
+++ b/eio/tests/fuzz.ml
@@ -148,10 +148,12 @@ end = struct
       | `Send len ->
         let available = String.length t.message - t.sent in
         let len = min len available in
-        let msg = Cstruct.of_string ~off:t.sent ~len t.message in
-        t.sent <- t.sent + len;
-        Log.info (fun f -> f "%a: sending %S" pp_dir t (Cstruct.to_string msg));
-        Eio.Flow.write sender [msg];
+        if len > 0 then (
+          let msg = Cstruct.of_string ~off:t.sent ~len t.message in
+          t.sent <- t.sent + len;
+          Log.info (fun f -> f "%a: sending %S" pp_dir t (Cstruct.to_string msg));
+          Eio.Flow.write sender [msg];
+        );
         aux ()
     in
     aux()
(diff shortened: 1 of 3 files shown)


eio/tests/fuzz.ml
```diff
@@ -287,13 +309,17 @@ let main client_message server_message actions =
...
-(*
-  Logs.(set_level (Some Info));
-  Logs.set_reporter (Logs_fmt.reporter ());
-  ignore action;
-  main "ping" "pong" [
-    Some (To_server, Send 5);
-    Some (To_client, Send 5);
-  ]
-*)
+  if true then
```
hannesm, line 312: sorry, I don't quite get this style of code? would you mind to explain the construct in more detail, and/or make it disappear. I've no clue in which cases the alternative should be executed (with manual code modification, likely).

eio/tls_eio.ml
```diff
@@ -29,7 +29,7 @@ module Raw = struct
       raise exn
 
   let write_t t cs =
-    try Flow.copy (Flow.cstruct_source [cs]) t.flow
+    try Flow.write t.flow [cs]
```
hannesm, line 32: is this really part of `improve fuzz test`? or a performance improvement? or a semantic change?

hannesm: It at least deserves a separate commit with a slightly more descriptive commit message.

eio/tests/fuzz.ml
talex5: If you get a crash and you want to reproduce it manually for testing (e.g. to try variations) you can input it here directly. Keeping it in an `if` makes it easier to switch between the two modes, means it still gets type-checked when disabled, and you don't get bothered by unused variable errors from dune.

eio/tls_eio.ml
talex5: It's mostly just a simplification. It could be an optimisation if the underlying flow implements `write` directly. Otherwise, `write` just falls back to doing a `copy`, as in the old code. I can remove it if you prefer.

eio/tests/fuzz.ml
hannesm: Do you think it is possible to use "cmdliner" and have this "true" hidden behind a flag?

eio/tls_eio.ml
hannesm: As said, a separate commit with a commit message that includes your above description would be sufficient for me. I've disliked myself in the past to sneak any library changes in "update tests" commits, not being able to properly track years later why something changed in one way or another.

bikallem:
My comment here - https://github.com/mirleft/ocaml-tls/pull/460#issuecomment-1363945250 - is applicable to this PR I believe.

talex5:
OK, I've now updated this PR to remove the change to `tls_eio.ml` so this PR only affects the fuzz tests (and I've updated the commit message to reflect that). We can make that change next time we're changing the code itself, to keep this PR clean.

> Do you think it is possible to use "cmdliner" and have this "true" hidden behind a flag?

It's a bit tricky, as crowbar likes to parse the command-line options itself (in an at_exit handler!). I've now removed the code you objected to. It was just a convenience.

@bikallem wrote:
> I believe we need to update the test to not generate send action after send_close action is generated

There is no "send_close". You mean `Shutdown_send`? That sends `Exit` to the sending fiber, which calls `Eio.Flow.shutdown` and then exits, so it already doesn't send anything after a shut down.

hannesm:
Thanks for your PR.

Issue #446, opened 2022-04-27
Invalid server extensions error on private grpc service


mbacarella:
Hi there. Thank you for the pure-OCaml TLS implementation.

I'm attempting to port code that uses `Async_ssl` over to use `Tls_async` instead.

The code accesses a gRPC service without public endpoints, so I need to provide a ca file copied from the server.  This flow actually works fine with `Async_ssl`, but fails with an "invalid server extensions" error when I provide the ca file to `Tls_async`.

I also tried `Tls_async` without the ca file and it connected fine. Maybe I'm configuring the client wrong?

```ocaml
    let host = t.grpc_host in
    let port = t.grpc_port in 
    (* ... *)
    let%bind res =
      let%bind authenticator =
        let module X509 = Tls_async.X509_async in
        let crls = None in
        let allowed_hashes = None in
        let ca_file = t.tls_cert in
        let param = X509.Authenticator.Param.ca_file ?crls ?allowed_hashes ca_file () in
        match%bind
          X509.Authenticator.Param.to_authenticator
            ~time:X509.Authenticator.Param.time
            param
        with
        | Ok a -> return a
        | Error e ->
          failwithf "X509_async.Authenticator: %s: %s" ca_file (Error.to_string_hum e) ()
      in
      let config = Tls.Config.client ~authenticator () in
      let hnp = Host_and_port.create ~host ~port in
      (* XXX: Create a Domain_name.t if host is not an IP address. *)
      let host = None in
      Tls_async.connect ~socket config (Tcp.Where_to_connect.of_host_and_port hnp) ~host
    in
    let%bind reader, writer =
      match res with
      | Ok ((_ : Tls_async.Session.t), r, w) -> return (r, w)
      | Error e ->
        eprintf "Error connecting to %s:%d: %s\n" host port (Error.to_string_hum e);
    in
    (* ... *)
 ```

Error message

```shell
Error connecting to localhost:10009: (Tls_failure
 (Error
  (AuthenticationFailure
    "leaf certificate X.509 certificate\
   \nversion 3\
   \nserial 101033868549717947339053637548736621515\
   \nalgorithm ECDSA-SHA256\
   \nissuer /O=lnd autogenerated cert/CN=localhost\
   \nvalid from 2021-12-03 00:21:36 +00:00 until 2023-01-28 00:21:36 +00:00\
   \nsubject /O=lnd autogenerated cert/CN=localhost\
[...]

PR #466, opened 2023-02-02
Option allow_ca_cert for buggy server certificates where CA=true

mbacarella:
Discussed previously in #446. This allows connecting to TLS endpoints that have set the CA=true flag on their server certificates. OpenSSL and other TLS libraries not based on OpenSSL allow this, so this is unfortunately a necessary evil.

As suggested by @hannesm, it's disabled by default and will print a warning if it comes up.

Usage is to simply create an authenticator with this parameter:

```ocaml
Authenticator.Param.ca_file ?crls ?allowed_hashes ~allow_ca_cert:true ca_file ()
```

This depends on the companion PR in https://github.com/mirleft/ocaml-x509/pull/160


hannesm:
settled with mirleft/ocaml-x509#161 (released as x509 0.16.3). Thanks for your work.

Issue #464, opened 2023-01-04
tls-eio: decide on improvements needed


talex5:
In https://github.com/mirleft/ocaml-tls/pull/460#issuecomment-1364121892, @hannesm commented:

> Since it seems both you @bikallem and @talex5 are working on TLS and EIO, maybe it'd make sense if both of you come to a common understanding what kind of semantics you'd like to have. [...] I already wonder whether I merged the tls-eio PR way too early.

I agree we're sending too many (half-baked) PRs. Let's first decide what behaviour we want and then write the code afterwards. But I think we really need @hannesm's input on this.

As I see it, there are 4 issues being discussed:

1. tls-eio may perform concurrent writes on a flow, and Eio does not specify the behaviour in this case.
2. Trying to use a TLS flow after it has failed produces confusing errors.
3. Trying to use a TLS flow after cancelling an operation is not supported and produces confusing errors.
4. Half-shutdown does not work, but the documentation doesn't mention this and the resulting behaviour is confusing.

As far as I can see, 1, 2 and 3 also affect tls-lwt (4 doesn't because the documentation doesn't claim to support it in the first place).

## Concurrent writes

Reading from a TLS flow may cause a write. This means that performing a read and a write of the TLS flow at the same time (which should be allowed) may cause overlapping writes on the underlying flow (which isn't).

I'm not sure why we haven't seen this problem in tls-lwt, which seems to be doing the same thing (`Lwt_cs.write_full` doesn't take a lock, even though it may do multiple writes on the FD).

This problem doesn't show up in the fuzz tests because the mock socket handles concurrent writes in order. But I tried modifying it to fail in the case of concurrent writes and it still didn't trigger. Do concurrent writes only happen on errors, perhaps?

In https://github.com/mirleft/ocaml-tls/pull/458#issuecomment-1346181930, @hannesm says:
> the other side-effecting pieces do not need a mutex due to careful design

I propose that Eio should explicitly document that the behaviour of concurrent writes is unspecified (see https://github.com/ocaml-multicore/eio/issues/387). It's possible that there's nothing to fix here, but I suspect we need some kind of lock when reporting errors at least. I can make a PR to do that, but I'd like to know whether this is really a problem and how to trigger it in the fuzz tests if so.

## Errors using a flow after failure

If an exception occurs while reading or writing the underlying flow, tls-eio moves into the `Error` state and all future operations fail with the same error.

[...]

PR #469, opened 2023-02-06
tls-lwt: do not catch out of memory exception

hannesm:
as discussed in #464 (thanks to @talex5 for raising this)

### DIFF
--- a/lwt/tls_lwt.ml
+++ b/lwt/tls_lwt.ml
@@ -43,16 +43,22 @@ module Unix = struct
   }
 
   let safely th =
-    Lwt.catch (fun () -> th >>= fun _ -> return_unit) (fun _ -> return_unit)
+    Lwt.catch
+      (fun () -> th >>= fun _ -> return_unit)
+      (function
+        | Out_of_memory -> raise Out_of_memory
+        | _ -> return_unit)
 
   let (read_t, write_t) =
     let recording_errors op t cs =
       Lwt.catch
         (fun () -> op t.fd cs)
-        (fun exn -> (match t.state with
-             | `Error _ | `Eof -> ()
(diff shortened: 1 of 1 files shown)


talex5:
Sorry, that wasn't what I meant.

Marking a flow as broken when you get `Out_of_memory` is a perfectly fine thing to do - it is very likely to be in a bad state and should not be used again.

Rather, I meant we want to be able to distinguish between:
- The system is currently out of memory (try dropping some caches), and
- This flow cannot be used because the system was previously out of memory (close it and open a new one).

`Out_of_memory` was just an example of an exception that is confusing if re-raised later.

hannesm:
Thanks for your comment @talex5, but I think the `Out_of_memory` exception is really exceptional, and the application should deal with it.There's nothing a library can do -- any further allocation (even for closing streams, etc.) may lead to subsequent failures. Read further on https://github.com/mirage/mirage/issues/1036 if you're interested in why it's bad to catch and hide `Out_of_memory`.

talex5:
I agree that libraries shouldn't be suppressing `Out_of_memory` (which the Mirage libraries did), but recording it and immediately raising it again is OK.

What I don't like is having special cases (e.g. what about `Stack_overflow` or `Assert_failure`?). The application should always get to see unexpected exceptions.

So, I think TLS should either store all unknown exceptions or none of them, and report it to the application in all cases.

The main tricky case is `read`, which may successfully read some data but then fail to write. Currently it reports success and then reports the write failure later.

> There's nothing a library can do -- any further allocation (even for closing streams, etc.) may lead to subsequent failures.

That's not a big problem. It's responding to an error in a complex operation (e.g. reading TLS data) by trying to perform a simpler one (e.g. marking the flow as failed). If that runs out of memory too, it will still get reported.

Issue #472, opened 2023-02-17
s-expression conversion: who's using it?


hannesm:
Hey,

is anyone actually using the to_sexp functions provided by the `deriving sexp` all over the place? I know the Tracing module uses it, but that's mostly useful for protocol debugging (when working on TLS itself).

Maybe @torinnd @mbacarella @talex5 @bikallem @dinosaure @cfcs have an opinion on this? If there isn't any use of it, I plan to remove the sexplib (and ppx) dependency cone in a future release.

Please let me know what you think.

talex5:
I don't think I've used it for anything, so fine with me to remove it.

bikallem:
I haven't had the need to use it so far.

reynir:
I guess I have indirectly used by enabling the `tls.tracing` log source in unikernels to inspect the data transfered. This could be achieved without sexp and I would be happy to have it removed as a dependency.

hannesm:
@reynir indeed. The data in / data out could still be done on a verbose log source. What would be more tricky would the state dump (and this was really useful when developing tls). Also, providing a text representation of Config.client/server records may be useful (though I don't think anyone is using that -- apart from some code in conduit (maybe @talex5 has memories thereof, I fail to understand the reasoning why there's sexp in conduit)).

hannesm:
And thanks @bikallem @talex5 for your answers.

talex5:
I don't know about conduit's use of sexp. @avsm added it very near the start of the Git history.

torinnd:
There are a couple spots in the async/ implementation where some other serialization will need to be used to generate error output, but I don't have any reservations about those changes being made. Seems fine to me!

hannesm:
@torinnd of course I'll ensure that error outputs (to the user) are retained in a nice way.

dinosaure:
Same as others, I don't really use it.

mbacarella:
I'm generally all-in on sexps for debugging but I admit I don't use them for tls.

avsm:
Just useful for debugging. I'm happy to reduce the dependency cone in the core library.  See uri (and uri-sexp) for a possible way to keep it, but in a separate package.

hannesm:
@avsm thanks for the suggestion. This could solve part of the problem, but not the interesting one: when there's a state change, report the new state, and also the types would need to be exposed publicly (i.e. state). So I think it is more trouble than worth it.

hannesm:
[...]

PR #473, opened 2023-02-22
remove ppx (cstruct and sexplib), removing sexplib converters

hannesm:
see discussion in #472 -- main purpose is to shave the dependency cone (and binary size reduces from 1.3 to 1.0 MB of tls.a)

still open for discussion (feedback, suggestions). will look into CI failures...


hannesm:
failing reverse dependencies:
- albatross.1.5.6 (failed: Unbound value Tls.Engine.string_of_failure)
- logs-syslog.0.3.2 (failed: Unbound value Tls.Engine.string_of_failure)
- conduit-mirage.6.2.0 (failed: Unbound value Tls.Config.client_of_sexp)
- mehari.0.2 (failed: Unbound module Tls.Core.Cert)
- jackline (failed: Unbound value Tls.Engine.string_of_failure)

PR #476, opened 2023-06-20
Upgrade JaneStreet async and core to v.0.16

tmcgilchrist:
Minor fixes to deprecated functions and function interfaces.

### DIFF
--- a/async/tls_async.mli
+++ b/async/tls_async.mli
@@ -55,4 +55,4 @@ val connect
       -> 'addr Tcp.Where_to_connect.t
       -> host:[ `host ] Domain_name.t option
       -> (Session.t * Reader.t * Writer.t) Deferred.Or_error.t)
-       Tcp.with_connect_options
+       Tcp.Aliases.with_connect_options
--- a/async/x509_async.ml
+++ b/async/x509_async.ml
@@ -9,7 +9,7 @@ let file_contents file =
 let load_all_in_directory ~directory ~f =
   let open Deferred.Or_error.Let_syntax in
   let%bind files = Deferred.Or_error.try_with (fun () -> Sys.ls_dir directory) in
-  Deferred.Or_error.List.map files ~f:(fun file ->
+  Deferred.Or_error.List.map ~how:`Sequential files ~f:(fun file ->
     let%bind contents = file_contents (directory ^/ file) in
     f ~contents)
 ;;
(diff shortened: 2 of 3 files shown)


hannesm:
Thanks for your effort. Asking @torinnd for a review since he has contributed the original async port of tls, and knows much more about async than me.

torinnd:
Sequential is indeed the right choice, and the other changes are also pretty reasonable. I'm happy with this change. Ocaml-CI, on the other hand, seems to have some experimental checks complaining, but I can't speak to that!

tmcgilchrist:
The linting check is suggesting these changes to opam files (it is ignorable) based on looking at the dune and opam files in the project:
```
tls-eio.opam: changes needed:
  "ptime" {>= "1.1.0"}                     [from eio]
  "logs" {with-test & >= "0.7.0"}          [from eio/tests]
tls-lwt.opam: changes needed:
  "ptime" {>= "1.1.0"}                     [from lwt]
```
Note: version numbers are just suggestions based on the currently installed version.
The other errors are for macOS ARM64, there might be a problem there with the EIO mdx tests, unreleased to this change. I don't have access to a suitable machine to check.

tmcgilchrist:
@hannesm an opam release of this would be helpful once you're happy with the changes.

hannesm:
> @hannesm an opam release of this would be helpful once you're happy with the changes.

would you mind to clarify two points:
- why would a release be helpful? I thought existing tls releases work with newer JS packages?
- why is there now a bound on 0.16.1? your commit says "include minor fix from ... to build on 5.1" -- but this just increases the lower bound of core to 0.16.1 (I find this highly suspicious, as I understand there is an issue to compile core 0.16.0 on ocaml 5.1, but this isn't anything that should affect tls-async... make the bounds for ocaml in core 0.16.0 / 0.16.1 tight enough, and you won't get that core 0.16.0 on ocaml 5.1)

hannesm:
oops, about the first point, I can see that any released tls-async actually doesn't work with the 0.16 series. still, please explain the bounds to core 0.16.1.

hannesm:
manually cherry-picked and pushed the first commit, thanks.

tmcgilchrist:
> please explain the bounds to core 0.16.1.

Core v0.16.0 has a compilation error on OCaml 5.1 alpha2, which is fixed in v0.16.1 https://github.com/ocaml/opam-repository/pull/24025.  I was trying to include two fixes in the one PR :-)

The first commit https://github.com/mirleft/ocaml-tls/pull/476/commits/f11ea05f5697be2d215172ed0d02bce77cf42c1e will solve things for the current released OCaml versions, thank you for the quick release. That unblocks my work on http2/grpc.  The second commit is necessary to support 5.1 in the future. I will create another PR to cover 5.1 support.

PR #477, opened 2023-07-04
Upgrade core to v0.16.1

tmcgilchrist:
Fixes bug in Core for OCaml 5.1.

### DIFF
--- a/tls-async.opam
+++ b/tls-async.opam
@@ -21,7 +21,7 @@ depends: [
   "ptime" {>= "0.8.1"}
   "async" {>= "v0.16"}
   "async_unix" {>= "v0.16"}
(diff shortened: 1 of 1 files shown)


hannesm:
would you mind to elaborate why my comment in the other issue does not apply?

> your commit says "include minor fix from ... to build on 5.1" -- but this just increases the lower bound of core to 0.16.1. This is highly suspicious, as I understand there is an issue to compile core 0.16.0 on ocaml 5.1, but this isn't anything that should affect tls-async.

This seems to be completely unrelated to this repository "fix core on ocaml 5.1".

tmcgilchrist:
Not required anymore. Thanks @hannesm

PR #479, opened 2023-09-19
Update to Eio 0.12

talex5:
(note: this looks long, but most of the changes are to the fuzz tests)

The changes to tls-eio.ml are:
- Add new `Tls_socket_closed` error. Writing to a closed connection previously reported end-of-file, which isn't quite right.
- Replace the object types with non-object types.
- Rename `read` to `single_read`, `write` to `single_write`, and `copy_from` to `copy`, to match the Eio 0.12 API.
- `copy` now uses an Eio helper that just calls `single_write`.

This removes all uses of objects from `tls-eio.ml`.

### DIFF
--- a/eio/tests/fuzz.ml
+++ b/eio/tests/fuzz.ml
@@ -296,7 +296,7 @@ let main client_message server_message quickstart actions =
       ~receiver:server_flow
       ~sender_closed:client_closed
       ~receiver_closed:server_closed
-      ~transmit:client_socket#transmit
+      ~transmit:(Mock_socket.transmit client_socket)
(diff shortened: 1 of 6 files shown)


hannesm:
I cut a release of tls-eio at https://github.com/ocaml/opam-repository/pull/24515 (since I suspect you'd like to use this rather sooner than later). Since the rest of tls* is not affected, I did not add these packages to the release (this should safe quite some CPU cycles for people not using tls-eio).

Enjoy the release.

talex5:
That's great - thanks!

PR #482, opened 2023-11-16
Support exporter secrets, as defined in RFC 5705 & 8446

hannesm:
//cc @reynir

### DIFF
--- a/lib/engine.ml
+++ b/lib/engine.ml
@@ -744,3 +744,36 @@ let epoch state =
   match epoch_of_hs state.handshake with
   | None -> `InitialEpoch
   | Some e -> `Epoch e
+
+let export_key_material (e : epoch_data) ?context label length =
+  match e.protocol_version with
+  | `TLS_1_3 ->
+    let hash =
+      let cipher = Option.get (Ciphersuite.ciphersuite_to_ciphersuite13 e.ciphersuite) in
+      Ciphersuite.hash13 cipher
+    in
+    let ems = e.exporter_master_secret in
+    let prk =
+      let ctx = Mirage_crypto.Hash.digest hash Cstruct.empty in
+      Handshake_crypto13.derive_secret_no_hash hash ems ~ctx label
+    in
+    let ctx = Option.(value ~default:Cstruct.empty (map Cstruct.of_string context)) in
+    Handshake_crypto13.derive_secret_no_hash
+      hash prk ~ctx:(Mirage_crypto.Hash.digest hash ctx)
+      ~length "exporter"
+  | #tls_before_13 as v ->
+    let seed =
+      let base =
+        match e.side with
+        | `Server -> Cstruct.append e.peer_random e.own_random
+        | `Client -> Cstruct.append e.own_random e.peer_random
+      in
+      match context with
+      | None -> base
+      | Some data ->
+        let len = Cstruct.create 2 in
+        Cstruct.BE.set_uint16 len 0 (String.length data);
+        Cstruct.concat [ base ; len ; Cstruct.of_string data ]
+    in
+    Handshake_crypto.pseudo_random_function v e.ciphersuite
+      length e.master_secret label seed
--- a/lib/handshake_client13.ml
+++ b/lib/handshake_client13.ml
@@ -181,8 +181,9 @@ let answer_finished state (session : session_data13) server_hs_secret client_hs_
   let myfin = Handshake_crypto13.finished hash client_hs_secret log in
   let mfin = Writer.assemble_handshake (Finished myfin) in
 
-  let resumption_secret = Handshake_crypto13.resumption session.master_secret  (log <+> mfin) in
-  let session = { session with resumption_secret ; client_app_secret ; server_app_secret } in
+  let exporter_master_secret = Handshake_crypto13.exporter session.master_secret log in
+  let resumption_secret = Handshake_crypto13.resumption session.master_secret (log <+> mfin) in
+  let session = { session with resumption_secret ; exporter_master_secret ; client_app_secret ; server_app_secret } in
   let machina = Client13 Established13 in
 
   Tracing.hs ~tag:"handshake-out" (Finished myfin);
--- a/lib/handshake_crypto.mli
+++ b/lib/handshake_crypto.mli
@@ -3,3 +3,6 @@ open State
 val derive_master_secret : Core.tls_before_13 -> session_data -> Cstruct.t -> Cstruct.t list -> Core.master_secret
 val initialise_crypto_ctx : Core.tls_before_13 -> session_data -> (crypto_context * crypto_context)
 val finished : Core.tls_before_13 -> Ciphersuite.ciphersuite -> Cstruct.t -> string -> Cstruct.t list -> Cstruct.t
+
+val pseudo_random_function : Core.tls_before_13 -> Ciphersuite.ciphersuite ->
+  int -> Cstruct.t -> string -> Cstruct.t -> Cstruct.t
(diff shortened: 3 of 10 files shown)


hannesm:
I tested with (or against) openssl s_server -keymatexport test -- though that's only the client side I tested (with various protocol versions).

lib/state.ml
```diff
@@ -411,9 +414,11 @@ let epoch_of_session server peer_name protocol_version = function
     let epoch : epoch_data = common_data_to_epoch session.common_session_data13 server peer_name in
     {
       epoch with
+      protocol_version       = protocol_version ;
```
hannesm, line 417: while here, I thought it'd be nice to actually include the TLS_1_3 in the epoch ;) (read: this was a bug before, and is fixed now)

lib/handshake_crypto.mli
```diff
@@ -3,3 +3,6 @@ open State
 val derive_master_secret : Core.tls_before_13 -> session_data -> Cstruct.t -> Cstruct.t list -> Core.master_secret
 val initialise_crypto_ctx : Core.tls_before_13 -> session_data -> (crypto_context * crypto_context)
 val finished : Core.tls_before_13 -> Ciphersuite.ciphersuite -> Cstruct.t -> string -> Cstruct.t list -> Cstruct.t
+
+val pseudo_random_function : Core.tls_before_13 -> Ciphersuite.ciphersuite ->
+  int -> Cstruct.t -> string -> Cstruct.t -> Cstruct.t
```
reynir, line 8: There are three `Cstruct.t`/`string` in a row so maybe a comment or labels would be nice.
```suggestion
(** [pseudo_random_function version cipher length secret label seed] *)
val pseudo_random_function : Core.tls_before_13 -> Ciphersuite.ciphersuite ->
  int -> Cstruct.t -> string -> Cstruct.t -> Cstruct.t
```

lib/handshake_server13.ml
```diff
@@ -342,7 +342,8 @@ let answer_client_hello ~hrr state ch raw =
 
       let session =
         let common_session_data13 = { session'.common_session_data13 with master_secret = master_secret.secret } in
-        { session' with common_session_data13 ; master_secret (* TODO ; exporter_secret *) }
```
reynir, line 345: Yay! A TODO removed! :-)

reynir (commented):
I think the TLS<1.3 looks convincing. For TLS 1.3 I need to dig through the relevant RFCs a bit more to understand it, but I don't see anything concerning :-)

lib/engine.ml
```diff
@@ -744,3 +744,36 @@ let epoch state =
...
+
+let export_key_material (e : epoch_data) ?context label length =
+  match e.protocol_version with
+  | `TLS_1_3 ->
+    let hash =
+      let cipher = Option.get (Ciphersuite.ciphersuite_to_ciphersuite13 e.ciphersuite) in
+      Ciphersuite.hash13 cipher
+    in
+    let ems = e.exporter_master_secret in
+    let prk =
```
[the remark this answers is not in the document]
hannesm: yes, so, there's (from RFC 8446):

```
       Derive-Secret(Secret, Label, Messages) =
            HKDF-Expand-Label(Secret, Label,
                              Transcript-Hash(Messages), Hash.length)
```

and now:
```OCaml
let derive_secret_no_hash hash prk ?length ?(ctx = Cstruct.empty) label =
  let length = match length with
    | None -> Mirage_crypto.Hash.digest_size hash
    | Some x -> x
  in
  let info = hkdflabel label ctx length in
  trace "prk" prk ;
  let key = Hkdf.expand ~hash ~prk ~info length in
  trace ("derive_secret: " ^ label) key ;
  key
```

so, derive-secret is indeed a `Hkdf.expand` with the secret (exporter_master_secret), and the "Messages" are "" -- this means we've to hash the empty string.

The only thing not discussed in here is the `hkdflabel` -- but that simply constructs the label out of label and context (well, and rfc 8446 again:
```
       struct {
           uint16 length = Length;
           opaque label<7..255> = "tls13 " + Label;
           opaque context<0..255> = Context;
       } HkdfLabel;
```

and that's what the OCaml function `hkdflabel` does as well. ;)

lib/engine.ml
```diff
@@ -744,3 +744,36 @@ let epoch state =
...
+      let ctx = Mirage_crypto.Hash.digest hash Cstruct.empty in
+      Handshake_crypto13.derive_secret_no_hash hash ems ~ctx label
+    in
+    let ctx = Option.(value ~default:Cstruct.empty (map Cstruct.of_string context)) in
+    Handshake_crypto13.derive_secret_no_hash
+      hash prk ~ctx:(Mirage_crypto.Hash.digest hash ctx)
+      ~length "exporter"
```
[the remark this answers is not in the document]
hannesm: short answer is yes, see the description above on that derive-secret is very similar (well, implemented in terms of hkdf-expand-label).

lib/handshake_server13.ml
hannesm: 🎉

lib/handshake_client13.ml
```diff
@@ -181,8 +181,9 @@ let answer_finished state (session : session_data13) server_hs_secret client_hs_
   let myfin = Handshake_crypto13.finished hash client_hs_secret log in
   let mfin = Writer.assemble_handshake (Finished myfin) in
 
-  let resumption_secret = Handshake_crypto13.resumption session.master_secret  (log <+> mfin) in
-  let session = { session with resumption_secret ; client_app_secret ; server_app_secret } in
+  let exporter_master_secret = Handshake_crypto13.exporter session.master_secret log in
+  let resumption_secret = Handshake_crypto13.resumption session.master_secret (log <+> mfin) in
```
hannesm, line 185: FWIW, RFC 8446 specifies the resumption secret to cover the client finished, while the exporter master secret only goes until server finished.

lib/handshake_server13.ml
```diff
@@ -342,7 +342,8 @@ let answer_client_hello ~hrr state ch raw =
...
+        let exporter_master_secret = Handshake_crypto13.exporter session.master_secret log in
```
hannesm, line 345: as mentioned above, the exporter secret only goes until the server finished, this is why we can nicely compute it here (and the resumption secret is computed at another place to include the client finished).

hannesm:
what is not part of this PR is the early_exporter_master_secret -- I thought I'd delay this until there's someone requesting it.

Neustradamus:
@hannesm: Good job!

Can you look for "tls-exporter"?
- https://datatracker.ietf.org/doc/html/rfc9266

Little details, to know easily:
- tls-unique for TLS =< 1.2
- tls-server-end-point
- tls-exporter for TLS = 1.3

Thanks in advance.

hannesm:
> @hannesm: Good job!
> 
> Can you look for "tls-exporter"?
> 
>     * https://datatracker.ietf.org/doc/html/rfc9266
> 
> 
> Little details, to know easily:
> 
>     * tls-unique for TLS =< 1.2
> 
>     * tls-server-end-point
> 
>     * tls-exporter for TLS = 1.3
> 
> 
> Thanks in advance.

Hello,

I'd be curious what you mean by `Can you look for "tls-exporter"?`? Would you mind to go into detail what your perspective is, and what you'd like to achieve?
Esp. what is missing from this PR in that direction? And what is the actual use case of 9266 (i.e. which application / protocol do you have in mind)?

Best,

Hannes

hannesm:
@Neustradamus if you need the RFC you mentioned above, please open a new issue with slightly more details: what and why should be supported by ocaml-tls. thanks a lot.

Neustradamus:
@hannesm: I will do a new ticket like I have done for a lot of others projects ^^

Already a good job done!

Linked to:
- https://github.com/scram-sasl/info/issues/1

PR #487, opened 2023-12-22
cleanups (mainly of mirage layer)

hannesm:
Best read commit-by-commit. ~~This is still work-in-progress, there'll be support for `shutdown` (to conform to mirage-flow 4) in a later commit.~~


mirage/tls_mirage.ml
```diff
@@ -37,16 +37,10 @@ module Make (F : Mirage_flow.S) = struct
...
 
-  let lift_write_result = function
-    | Ok ()   -> `Ok ()
-    | Error e -> `Error (`Write e)
-
   let check_write flow f_res =
-    let res = lift_write_result f_res in
-    ( match flow.state, res with
-      | `Active _, (`Eof | `Error _ as e) ->
-          flow.state <- e ; FLOW.close flow.flow
```
dinosaure, line 48: I think this diff is a part for the `shutdown` PR, I'm right?

dinosaure (approved):
Seems right even if I think that the deletion of `Flow‧close` should be done into the `shutdown` (for a better clarity in my opinion but that's fine).

mirage/tls_mirage.mli
```diff
@@ -20,6 +20,9 @@ module Make (F : Mirage_flow.S) : sig
     with type error := error
      and type write_error := write_error
 
+  (** [underlying t] returns the underlying flow. This is useful to extract
+      information such as [src] and [dst] of that flow. *)
+  val underlying : flow -> FLOW.flow
```
dinosaure, line 25: :tada:

mirage/tls_mirage.ml
hannesm: no -- well, the purpose here is to prepare for shutdown. A first step is to not `close` the underlying flow on arbitrary errors -- so the caller has to close the flow on error conditions.

lib/packet.ml
```diff
@@ -14,28 +14,24 @@ type content_type =
...
   | HANDSHAKE -> 22
   | APPLICATION_DATA -> 23
-  | HEARTBEAT -> 24
 and int_to_content_type = function
   | 20 -> Some CHANGE_CIPHER_SPEC
   | 21 -> Some ALERT
   | 22 -> Some HANDSHAKE
   | 23 -> Some APPLICATION_DATA
-  | 24 -> Some HEARTBEAT
   | _ -> None
```
reynir, line 28: Now this is responsible for erroring out on `HEARTBEAT`?

reynir (approved):

lib/packet.ml
hannesm: Yes, exactly. The `int_to_content_type` is used by `Reader.parse_record`, which errors out with `UnknownContent _` - which leads (in `Engine.separate_records` to `` Error (`Fatal (`UnknownContentType _)) ``.

There is still an open question about error semantics -- which cases are fatal, which are not. For the heartbeat, this requires a negotiation in the client/server hello (which we do not support) -- thus anyone sending us a heartbeat packet violates the protocol.

hannesm:
> Seems right even if I think that the deletion of `Flow‧close` should be done into the `shutdown` (for a better clarity in my opinion but that's fine).

Ok, I removed that close commit from this PR.

PR #488, opened 2024-01-04
Shutdown

hannesm:
this is still work in progress. The idea is to track in engine the status of (half-)closed connections - and implement the revised close_notify semantics (i.e. not replying immediately with a close_notify).

There are still some checks missing (e.g. for reneg/rekey/send_application_data), also should the effectful layers inform the engine when a read or write failed (unclear, tendency towards no)?


hannesm:
In this PR, the changes to tls-lwt are done now. Curious to hear feedback if there's any.

hannesm:
now the tls-mirage bits as well use the new API and provide the mirage-flow 4 interface (shutdown).

reviews / suggestions would be welcome (esp. @reynir @dinosaure who had great reviews for the awa-ssh PR that is similar to this here)

also, if there is anyone eager to adjust async (@torinnd @bcc32 @tmcgilchrist) and eio (@talex5 @bikallem) bits, that'd be great. I'm open for discussions about the changes to the Engine API above.

dinosaure:
I actually tested this PR with Miou and `httpaf` and it seems to work pretty well at a glance :+1:.

talex5:
Thanks - this is great!

I pushed updates for tls-eio to https://github.com/talex5/ocaml-tls/commits/shutdown/ (needs review).

The first commit just ports the changes to tls-lwt directly to tls-eio. The second removes the work-arounds for missing half-shutdown in the fuzz tests, which are passing so far (after 5 minutes of afl-fuzz).

lwt/tls_lwt.ml
```diff
@@ -56,43 +75,47 @@ module Unix = struct
...
-    | `Active _ ->
-        read_t t t.recv_buf >>= fun n ->
-        match (t.state, n) with
-        | (`Active _  , 0) -> t.state <- `Eof ; return `Eof
-        | (`Active tls, n) -> handle tls (Cstruct.sub t.recv_buf 0 n)
-        | (`Error e, _)    -> fail e
-        | (`Eof, _)        -> return `Eof
+    | `Error e -> Lwt.reraise e
+    | `Closed
+    | `Read_closed _ -> Lwt.return `Eof
```
dinosaure, line 107: If I look into `tls_mirage.ml`, the `` `Closed`` state produces a `` `Write `Closed`` where `tls_lwt.ml` produces a simple `` `Eof``. I'm in favor for the `tls_lwt.ml` version where a `ECONNRESET` can appear and should be handled as an `` `Eof``.

hannesm: It is tricky -- so in tls_lwt, a read when `` `Read_closed `` or `` `Closed `` is the internal state, and `` `Eof `` is signalled. If `writev` (or `write`) is called with `` `Closed `` as internal state, an exception is raised.

In mirage, I pushed 3ed1d0e to adhere to this behaviour. This is similar to how awa-ssh does it. :)

Thanks for your review.

hannesm:
CI failures are in the eio, I don't quite understand the issues there. @talex5 I see two options: you open your changes as a PR to this repository, I merge and release -- or I release without the eio part. any preference?

talex5:
I think you can just `git pull` my branch into yours to fix it.

talex5:
Or I can open a PR against this branch if you prefer.  The logs for the fuzz failure on your branch look like this:

```
+dispatch_commands: done
+client-to-server: receiver ready
+server-to-client: sender ready
+server-to-client: shutdown send (Tls level)
+server: wrote 24 bytes to network
+client: read 24 bytes from network
+server-to-client: recv thread done (got EOF)
+client-to-server: shutdown send (Tls level)
Fatal error: exception Eio_mock__Backend.Deadlock_detected
```

But I didn't investigate since I assumed it wouldn't work without the changes anyway.

hannesm:
if you open a PR against my branch, that'd be excellent

talex5:
OK: https://github.com/hannesm/ocaml-tls/pull/5

PR #492, opened 2024-04-24
Document why Tls.Engine.send_application_data is None

reynir:
As discussed in https://github.com/robur-coop/miragevpn/pull/234/files/4f553471dc81e03790ead342caba5e50ac05fd71#r1576424941


hannesm:
thanks!

PR #493, opened 2024-04-29
use bytes/string instead of cstruct

hannesm:
still needs to adapt the other effectful layers (mirage/eio/async)


hannesm:
see #497

PR #494, opened 2024-05-01
Add the Miou implementation of TLS

dinosaure:
This PR provides an implementation of `tls` with [miou][miou]. It provides also a test which checks randomly different actions between a client and a server. The package is actually tested with [httpcats][httpcats] (for http/1.1 and h2) and, if we want to take the opportunity to do some HTTP requests in parallel, this [patch][mirage-crypto] about `mirage-crypto` is needed - but it is not necessary if we only use `Miou.call_cc` which spawns a task concurrently.

## The fuzzer

The fuzzer tests the behavior of `Tls_miou` when we use these actions:
```ocaml
type operation =
  | Send of string
  | Recv of int
  | Shutdown of [ `read | `write ]
  | Close
  | Noop
```

It does not test the TLS implementation (if we are able to correctly encrypt/decrypt the session). These actions are randomly chosen by the fuzzer. The idea is to produce a suit of actions for a client and a server. Note that we don't generate a situation where both read or both write "at the same time". The generation has 4 points where we are able to `close`, `shutdown(recv)` or `shutdown(write)`. An example is:
```
client: [send(64); noop; recv(64); shutdown(send); recv(64); noop; noop; close; noop; close]
server: [recv(64); noop; send(64); shutdown(send); send(64); close; noop; close; noop; close]
```

Then, we launch a TLS server and start a new session with a client. Both do actions and fill a buffer. We catch all possible errors and terminate properly the server and the client. Finally, we compare what the server received and what the client received with an _oracle_. The oracle is the `compile` function. The rules are simple:
- we record when the client and the server shutdown/close the connection
- if the client/server want to send/recv something, we check the status of the connection. We continue if everything is normal (none of them shutdown/close their sides) or we directly return the final result (in our implementation, we should raise an exception `End_of_file` or `Closed_by_peer`).

## The error cases

Actually, from an UNIX socket, `read` can raises `ECONNRESET` and `write` can raise `EPIPE` and `ECONNRESET`. For our perspective, these cases tell us that the connection is closed (on a specific side or totally). We handle these cases and transfer the information to the internal TLS state.

Another issues encountered is about `read` and `close`. Our fuzzer uses a UNIX socket - we can not rely on the throughput of the network. It can arrives that the server get the whole data and a close-notify before to do anything. In this situation, we must returns what we got and consider, afterwards, the connection closed even if the TLS state is actually `` `Closed``.

Finally, the previous point enforces the fact that we must returns what the client/server received even if the connection was closed (or `shutdown(recv)`). However, the user can also decide to `close` or `shutdown(recv)` the connection **before** to read anything. We must drop pending _application-data_ in such situation.

## Parallelism

Currently, as said in the introduction, we must use a patch for `mirage-crypto` to be able to launch several parallel TLS sessions. The `mirage-crypto-rng-miou` is also needed if we don't want any data-race conditions when we generate a random value. The test actually uses only one domain and `mirage-crypto-rng.unix` according to the current situation of `mirage-crypto`.

[miou]: https://github.com/robur-coop/miou
[httpcats]: https://github.com/robur-coop/httpcats
[mirage-crypto]: https://github.com/mirage/mirage-crypto/pull/186


dinosaure:
A new branch `dinosaure/ocaml-tls#miou-string` exists which is this PR on top of our changes about `Cstruct.t -> string` but I will propose it when the `Cstruct.t -> string` PR will be merge. The code does not change to much. Let me know when everything will be good about `tls` to propose then a rebased version of this PR.

hannesm:
I thought we'll do the cstruct -> string jump in mirage-crypto with a release soon, and then adapt x509 etc. as well, and get this PR merged when we also have tls using string/bytes. Or should this be merged earlier and released (I thought it'd depend on mirage-crypto-rng-miou).

dinosaure:
We should merge this after the Cstruct.t - > string PR. It does not require as is mirage-crypto-rng-miou-unix but it requires fixes about parallelism to work nicely then (so it requires a release of mirage-crypto which includes the Cstruct.t -> string move).

hannesm:
This can now that #497 is merged be rebased on the current main branch.

hannesm:
superseeded by #503

PR #495, opened 2024-05-10
Fix eporter secret computations

hannesm:
//cc @reynir


hannesm:
I tested this with miragevpn -- before this PR, our server couldn't use tls-ekm with our client - with this PR, it can! :)

PR #496, opened 2024-05-10
support channel bindings (RFC 5929, 9266)

hannesm:
fixes #484

PR #497, opened 2024-06-27
TLS without `cstruct`

dinosaure:
#493 rebased to include #495 & #496 plus an update of `tls-async`. The only remaining package about the `no-cstruct` move is `tls-eio`.


art-w:
In case it helps, this patch https://github.com/mirleft/ocaml-tls/commit/05b203ec58569e7b1cb55d29d43a2c357f260603 should provide the missing `tls-eio` :)

hannesm:
this is great. I guess we need to rebase and wait for the x509 release. (plus then adapt lower bounds)

art-w:
Nice rebase! The eio patch and a fix for the latest X509 breaking change (renaming of Authenticator functions) can be found at https://github.com/art-w/ocaml-tls/commits/no-cstruct/

hannesm:
@dinosaure would you mind to pick @art-w commits here?

async/io.ml
```diff
@@ -73,18 +73,17 @@ module Make (Fd : Fd) : S with module Fd := Fd = struct
        | Active _, `Eof ->
          t.state <- Eof;
          return `Eof
-       | Active tls, `Ok n -> handle tls (Cstruct.sub t.recv_buf 0 n)
+       | Active tls, `Ok n -> handle tls (Stdlib.Bytes.sub_string t.recv_buf 0 n)
```
hannesm, line 76:
```suggestion
       | Active tls, `Ok n -> handle tls (Bytes.sub_string t.recv_buf 0 n)
```

async/io.ml
```diff
@@ -73,18 +73,17 @@ module Make (Fd : Fd) : S with module Fd := Fd = struct
...
   let rec read t buf =
     let writeout res =
-      let open Cstruct in
-      let rlen = length res in
-      let n = min (length buf) rlen in
-      blit res 0 buf 0 n;
-      t.linger <- (if n < rlen then Some (sub res n (rlen - n)) else None);
+      let rlen = String.length res in
+      let n = min (Bytes.length buf) rlen in
+      Stdlib.Bytes.blit_string res 0 buf 0 n;
```
hannesm, line 85:
```suggestion
      Bytes.blit_string res 0 buf 0 n;
```

async/io.ml
```diff
@@ -73,18 +73,17 @@ module Make (Fd : Fd) : S with module Fd := Fd = struct
...
+      t.linger <- (if n < rlen then Some (Stdlib.String.sub res n (rlen - n)) else None);
```
hannesm, line 86:
```suggestion
      t.linger <- (if n < rlen then Some (String.sub res n (rlen - n)) else None);
```

lib/core.ml
```diff
@@ -3,12 +3,14 @@
...
 
-let (<+>) = Cstruct.append
-
 let ( let* ) = Result.bind
 
 let guard p e = if p then Ok () else Error e
 
+let split_str ?(start = 0) str off =
+  String.sub str start off,
+  String.sub str (start + off) (String.length str - off - start)
```
hannesm, line 12: it's likely I introduced this function, but from our learnings in MirageVPN, I guess we should re-visit the allocation and ownership stuff in this PR. The Strong.sub is super-expensive, and we should try to allocate the least possible.

This doesn't need to be in this PR, but before a release. We should as well benchmark the amount of handshakes and throughput before and after this change (with varying block sizes in send()). Not sure whether we have a suitable executable for this right now. Best to compare to OpenSSL on the same machine.

lib/crypto.ml
```diff
@@ -69,76 +63,69 @@ module Ciphers = struct
 end
 
 let sequence_buf seq =
-  let open Cstruct in
-  let buf = create 8 in
-  BE.set_uint64 buf 0 seq ;
-  buf
+  let buf = Bytes.create 8 in
+  Bytes.set_int64_be buf 0 seq ;
+  Bytes.unsafe_to_string buf
```
hannesm, line 68: allocation-wise likely a bad idea

lib/crypto.ml
```diff
@@ -69,76 +63,69 @@ module Ciphers = struct
...
-    let s = sequence_buf seq in
-    let pad = Cstruct.create (l - 8) in
-    pad <+> s
+    let l = String.length nonce in
+    let buf = Bytes.make l '\x00' in
+    Bytes.set_int64_be buf (l - 8) seq;
+    Bytes.unsafe_to_string buf
   in
-  Uncommon.Cs.xor nonce s
+  Uncommon.xor nonce s
```
hannesm, line 77: same, likely better withput allocations, and xor_into

lib/handshake_client.ml
```diff
@@ -424,9 +427,9 @@ let answer_server_finished state (session : session_data) client_verify fin log
   let computed =
     Handshake_crypto.finished (state_version state) session.ciphersuite session.common_session_data.master_secret "server finished" log
   in
-  let* () = guard (Cstruct.equal computed fin) (`Fatal `BadFinished) in
+  let* () = guard (String.equal computed fin) (`Fatal `BadFinished) in
```
hannesm, line 430: do we need eqaf here?

lib/handshake_client.ml
```diff
@@ -98,7 +98,7 @@ let common_server_hello_validation config reneg (sh : server_hello) (ch : client
 
 let common_server_hello_machina state (sh : server_hello) (ch : client_hello) raw log =
   let cipher = sh.ciphersuite in
-  let session_id = match sh.sessionid with None -> Cstruct.create 0 | Some x -> x in
+  let session_id = match sh.sessionid with None -> "" | Some x -> x in
```
hannesm, line 101:
```suggestion
  let session_id = Option.value ~default:"" sh.sessionid in
```

lib/handshake_client.ml
```diff
@@ -141,9 +141,9 @@ let answer_server_hello state (ch : client_hello) sh secrets raw log =
 
   let* () =
     if max_protocol_version state.config.protocol_versions = `TLS_1_3 then
-      let piece = Cstruct.sub sh.server_random 24 8 in
-      let* () = guard (not (Cstruct.equal Packet.downgrade12 piece)) (`Fatal `Downgrade12) in
-      guard (not (Cstruct.equal Packet.downgrade11 piece)) (`Fatal `Downgrade11)
+      let piece = String.sub sh.server_random 24 8 in
```
hannesm, line 144: allocates, maybe we have a function in eqaf or define our own to have off and len params for eq?

lib/handshake_client.ml
```diff
@@ -225,7 +225,7 @@ let answer_certificate_RSA state (session : session_data) cs raw log =
     | x -> Error (`Fatal (`NoVersions [ x ])) (* TODO: get rid of this... *)
   in
   let ver = Writer.assemble_protocol_version version in
-  let premaster = ver <+> Mirage_crypto_rng.generate 46 in
+  let premaster = ver ^ Mirage_crypto_rng.generate 46 in
```
hannesm, line 228: what we want to do: allocate 48 bytes, encode version into the first two bytes, generate_into in the remaining

lib/handshake_client13.ml
```diff
@@ -20,24 +20,27 @@ let answer_server_hello state ch (sh : server_hello) secrets raw log =
...
-          | None, _ | _, None -> Ok (Cstruct.create hlen, false)
+          | None, _ | _, None -> Ok (String.make hlen '\x00', false)
           | Some idx, Some (psk, _epoch) ->
             let* () = guard (idx = 0) (`Fatal `InvalidServerHello) in
             Ok (psk.secret, true)
         in
         let early_secret = Handshake_crypto13.(derive (empty cipher) psk) in
         let hs_secret = Handshake_crypto13.derive early_secret shared in
-        let log = log <+> raw in
+        let log = log ^ raw in
```
hannesm, line 39: if I remember, we should be able to use a hash state since we never need the entire log data. this is for sure a separate PR.

lib/handshake_common.ml
```diff
@@ -332,15 +330,14 @@ let server_hello_valid (sh : server_hello) =
...
   let ctx =
-    let stop = Cstruct.create 1 (* trailing 0 byte *) in
+    let stop = String.make 1 '\x00' (* trailing 0 byte *) in
     match context_string with
     | None -> stop
-    | Some x -> Cstruct.of_string x <+> stop
+    | Some x -> x ^ stop
   in
-  prefix <+> ctx
+  prefix ^ ctx
```
hannesm, line 340: allocate less

lib/handshake_crypto.ml
```diff
@@ -24,56 +20,55 @@ let prf_mac = function
...
-  let c_mac, rt0 = split buf mac in
-  let s_mac, rt1 = split rt0 mac in
-  let c_key, rt2 = split rt1 key in
-  let s_key, rt3 = split rt2 key in
-  let c_iv , s_iv = split rt3 iv
+  let c_mac, rt0 = Core.split_str buf mac in
+  let s_mac, rt1 = Core.split_str rt0 mac in
+  let c_key, rt2 = Core.split_str rt1 key in
+  let s_key, rt3 = Core.split_str rt2 key in
+  let c_iv , s_iv = Core.split_str rt3 iv
```
hannesm, line 58: this is 10 string.sub. we can easily have fewer by using the offset and string.sub directly

lib/handshake_crypto13.ml
```diff
@@ -7,13 +7,13 @@ let cdiv (x : int) (y : int) =
...
-  padding <+> msg
+  let padding = String.make (bytes - String.length msg) '\x00' in
+  padding ^ msg
 
 let not_all_zero r =
   let* cs = r in
-  let all_zero = Cstruct.create (Cstruct.length cs) in
-  if Cstruct.equal all_zero cs then
+  let all_zero = String.make (String.length cs) '\x00' in
+  if String.equal all_zero cs then
```
hannesm, line 16: I'm sure we can compare whether everything is 0 without allocating...

hannesm:
I think this is already pretty big, and we should work towards merging it. From my point of view, what is important:

- a benchmark comparing tls_version, ciphersuite, block size of write() calls of latest release and this PR (as a separate PR, to be merged before this PR
- have as well a script for that matrix using OpenSSL

Then we can compare and merge this PR and prepare performance improvements in subsequent changesets (e.g. allocate less). We as well have these decrypt/encrypt_into functions in mirage-crypto that we should use.

If someone would be eager to develop the benchmark, that'd speed up our release schedule:)

tls.opam
```diff
@@ -16,19 +16,19 @@ build: [
...
+  "mirage-crypto" {>= "1.0.0"}
+  "mirage-crypto-ec" {>= "1.0.0"}
+  "mirage-crypto-pk"
+  "mirage-crypto-rng" {>= "1.0.0"}
+  "x509" {>= "1.0.0"}
   "domain-name" {>= "0.3.0"}
   "fmt" {>= "0.8.7"}
-  "cstruct-unix" {with-test & >= "3.0.0"}
   "ounit2" {with-test & >= "2.2.0"}
   "hkdf" {< "2.0.0"}
```
hannesm, line 27:
```suggestion
  "hkdf" {>= "2.0.0"}
```

tls.opam
```diff
@@ -16,19 +16,19 @@ build: [
...
   "logs"
   "ipaddr"
+  "ohex" {>= "0.2.0"}
+  "digestif"
```
hannesm, line 31: likely needs some lower bound

tls-eio.opam
```diff
@@ -19,7 +19,6 @@ depends: [
   "tls" {= version}
   "mirage-crypto-rng" {>= "0.11.2" & < "1.0.0"}
```
hannesm, line 20:
```suggestion
  "mirage-crypto-rng" {>= "1.0.0"}
```

tls-eio.opam
```diff
@@ -19,7 +19,6 @@ depends: [
...
   "mirage-crypto-rng-eio" {with-test & >= "0.11.2" & < "1.0.0"}
```
hannesm, line 21:
```suggestion
  "mirage-crypto-rng-eio" {with-test & >= "1.0.0"}
```

tls-lwt.opam
```diff
@@ -18,7 +18,6 @@ depends: [
   "dune" {>= "3.0"}
   "tls" {= version}
   "mirage-crypto-rng-lwt" {>= "0.11.0" & < "1.0.0"}
```
hannesm, line 20:
```suggestion
  "mirage-crypto-rng-lwt" {>= "1.0.0"}
```

tls.opam
```diff
@@ -16,19 +16,19 @@ build: [
...
   "dune" {>= "3.0"}
-  "cstruct" {>= "6.0.0"}
-  "mirage-crypto" {>= "0.11.0" & < "1.0.0"}
-  "mirage-crypto-ec" {>= "0.10.0" & < "1.0.0"}
-  "mirage-crypto-pk" {< "1.0.0"}
-  "mirage-crypto-rng" {>= "0.8.0" & < "1.0.0"}
-  "x509" {>= "0.15.0"}
+  "mirage-crypto" {>= "1.0.0"}
+  "mirage-crypto-ec" {>= "1.0.0"}
+  "mirage-crypto-pk"
```
hannesm, line 21:
```suggestion
  "mirage-crypto-pk" {>= "1.0.0"}
```

hannesm:
I don't understand the tls-eio mdx failures... I see two options: remove it or someone who's into that stuff proposing a fix /cc @talex5 (who developed these bits)

lib/crypto.ml
dinosaure: What do you mean? We should use `Bytes.to_string`?

lib/handshake_client.ml
dinosaure: I did a try about implementations of `String.sub_equal` (which does not guarantee the constant time as `eqaf`). My benchmark is available here: https://gist.github.com/dinosaure/7ef54e9431217435e3b81df890192b72. From my results (on AMD), the "naive" implementation (the Oracle one) is fast enough for the worst case. For readability, I will consider this implementation for `ocaml-tls` (into the `lib/utils.ml` file) than the other.

lib/handshake_crypto13.ml
dinosaure: I decided to make a loop which raise an exception when we have something different than `\x00` and returns `Ok`. A double check of my implementation will be great.

talex5:
> I don't understand the tls-eio mdx failures

It doesn't seem to have much to do with tls-eio or mdx; just trying to load tls in utop no longer works:

```
$ utop
utop # #require "tls";;
Error: Reference to undefined compilation unit `Digestif'
Hint: This means that the interface of a module is loaded, but its implementation is not.
      Did you mean to load a compiled implementation of the module 
      using #load or by passing it as an argument to the toplevel?
```

hannesm:
> > I don't understand the tls-eio mdx failures
> 
> It doesn't seem to have much to do with tls-eio or mdx; just trying to load tls in utop no longer works:
> 
> ```
> $ utop
> utop # #require "tls";;
> Error: Reference to undefined compilation unit `Digestif'
> Hint: This means that the interface of a module is loaded, but its implementation is not.
>       Did you mean to load a compiled implementation of the module 
>       using #load or by passing it as an argument to the toplevel?
> ```

thanks for your investigation. @dinosaure is it known and is there a workaround for digestif not being loadable into ocaml toplevel (due to virtual library?)?

lib/crypto.ml
hannesm: no, the entire function and allocation discipline should be different. better to allocate a bytes once, and then filling it & passing offsets along. this requires, as mentioned, a pretty deep rewriting -- which brings the suggestion to first work on benchmarks (for the latest release - e2e, maybe similar to what @reynir worked on for MirageVPN -- but would be great if cstruct allocation could be part of the benchmark), to be able to compare with the changes in here and future changes.

dinosaure: > better to allocate a bytes once

Not sure to understand again, the content of this buffer actually depends on the given argument (`seq`). We probably can pre-allocate a set of `sequence_buf` and use them then but not sure this is what you want. If I think in this way, it's related to the multicore usage than can be done with `ocaml-tls`. In such context, "allocate a bytes once" (globally? per-domain?) can be risky.

dinosaure:
> thanks for your investigation. @dinosaure is it known and is there a workaround for digestif not being loadable into ocaml toplevel (due to virtual library?)?

You need to load an implementation of `digestif`. `digestif.c` or `digestif.ocaml`. But in my opinion, maybe we should reconsider `mdx` in the tests. I've already seen `mdx` related errors in the release process of some libraries and it's a burden that shouldn't fall on the `ocaml-tls` maintainers.

lib/crypto.ml
hannesm: oh, sorry for the misunderstanding. my intention is: we get some data to be sent out. now, instead of allocating a header for crypto, a header for tls, etc. - let's allocate once a buffer where the encrypted payload together with the protocol header has space. thus we won't need any other allocations, neither a concat or append. this is all local to a send_data operation, not global/per domain.

nothing for this pr, though.

lib/handshake_client.ml
dinosaure: Actually, and according to my benchmark, it seems that an `equal_sub` which uses `String.sub` or an `equal_sub` which tries to iterate into given `strings` have the same performance. My conclusion is: even if we do a `String.sub` here, try to avoid this `String.sub` and try to directly read bytes and compare them is not faster than the `String.sub`...

dinosaure:
This PR is in a good shape.

We can still imagine some optimizations and avoid certain allocations, but I'd still like to point out that allocating a small string (3 processor instructions) is still much more optimized than allocating a `Cstruct.t` (`malloc`), and that this PR is still a non-negligible gain overall.

I'd also like to point out that preferring to read a buffer rather than a `String.sub` (which involves allocation + copying) isn't necessarily faster (according to our micro-benchmark about `sub_equal`). This may lead us to the conclusion that writing to a buffer is perhaps the area we should be concentrating on.

A transitive breaking-change, however, has not been seen. Without `dune`, users will have to specify which `digestif` implementation they want if they want to use `tls` now.

hannesm:
force-pushed on top of #501

hannesm:
at least the tests/feedback.exe now works here as well.

still need to investigate/improve the benchmarks, and then we're ready to go \o/

hannesm:
force-pushed on top of main (#500 was merged), and fixed speed.ml

PR #498, opened 2024-07-15
Update upper bounds on mirage-crypo and hkdf

MisterDA:
- mirage-crypto* < 1.0.0
- hkdf < 2.0.0

The upper-bounds are taken from the opam-repository.

### DIFF
--- a/tls-async.opam
+++ b/tls-async.opam
@@ -24,7 +24,7 @@ depends: [
   "core" {>= "v0.16"}
   "cstruct-async"
   "ppx_jane" {>= "v0.16"}
-  "mirage-crypto-rng-async"
+  "mirage-crypto-rng-async" {< "1.0.0"}
 ]
 tags: [ "org:mirage"]
 synopsis: "Transport Layer Security purely in OCaml, Async layer"
--- a/tls-eio.opam
+++ b/tls-eio.opam
@@ -17,8 +17,8 @@ depends: [
   "ocaml" {>= "5.0.0"}
   "dune" {>= "3.0"}
   "tls" {= version}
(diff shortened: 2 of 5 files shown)


hannesm:
Not sure it is worth it. I'm on vacation, but have plans in August to move on with #497 (on e we have x509 released).

Is your desire to have fixes / improvements in the current main branch? Or why did you submit this PR?

MisterDA:
The normalization helps when diff'ing the `opam` file from this repo and how opam _prints_ the package metadata with `opam show --raw <pkg>`.

art-w:
Thanks for removing the normalization! It was a bit aggressive to reformat those files and break git-blame: even the `opam-repository` doesn't [normalize them](https://github.com/mirleft/ocaml-tls/blob/main/tls.opam), and it's more important to be able to diff with those than with the pretty-printed output of `opam show` :)

hannesm:
> The normalization helps when diff'ing the `opam` file from this repo and how opam _prints_ the package metadata with `opam show --raw <pkg>`.

not sure I follow, where/when is `opam show --raw <pkg>` used?

MisterDA:
> > The normalization helps when diff'ing the `opam` file from this repo and how opam _prints_ the package metadata with `opam show --raw <pkg>`.
> 
> not sure I follow, where/when is `opam show --raw <pkg>` used?

It's what I use to see the diff between the `*.opam` files in repositories and in opam-repository to "backport" the bounds. If the files are not normalized the diff is too noisy. I usually go with something like `diff <(opam lint --normalise tls.opam) <(opam show --raw tls)`.

PR #500, opened 2024-08-06
Add a benchmark for throughput and handshakes using bechamel

hannesm:
This is based on @reynir's work in MirageVPN

//cc @dinosaure


hannesm:
We could add:
- handshake benchmarks (taking different kex into account, FF-DHE, ECDSA, 25519) - also with different key types (at the moment RSA is used)
- different block sizes (at the moment it is 1024 bytes)
- server side benchmarks (send, receive, handshake)

I don't quite understand the output (would be great to have a "throughput in MB/s" and "handshake in number of handshakes per second").

This is a start for conducting some stats with the latest release in order to measure and investigate #497 performance.

Happy to get feedback on the concrete implementation and what matrix would make sense.

hannesm:
@dinosaure I pushed a commit to also conduct the handshake/s in throughput.exe. the same commit extends some more ciphersuites, and finally outputs markdown tables. WDYT?

dinosaure (approved):
We probably can delete the benchmark with `bechamel`. For the handshake, `bechamel` can be a little more “accurate” in the sense that it tries to execute the handshake several times and tries to guess the factor that determines the execution time between 10, 20, ..., 100 handshakes. It does this by analyzing the results to remove “noise”, as long as it has no influence on the coefficient of determination.

The proposed `throughput` framework does not do this, and the results can therefore be influenced by “noise” and influence our interpretation. So, as you wish!

hannesm:
Thanks for your reply @dinosaure. I'm in favour of having a single module with the benchmarks. We can later always revise.

Indeed, there's noise in the benchmarks. I suspect that we'll at some point get towards less noise :)

PR #501, opened 2024-08-19
tests/feedback: fix for TLS 1.3, run as test

hannesm:
- fix feedback to finish the handshake until server and client are ready Previously, it was only waited until client is ready. This in not sufficient for the TLS 1.3 state machine. In TLS < 1.3, the server sends a final "Finished" message, while in TLS 1.3, the client sends the final "Finished" message. In feedback operations, this was dropped, leading to MAC mismatch.
- tests/server.{key,pem} are symbolic links to ../certificates/server.{key,pem}
- tests/feedback.exe is run as test now
- tests/feedback.exe uses cmdliner and logs now

//cc @dinosaure this fixes feedback for the current main branch. I'll soon rebase the #497 PR on top of this to figure out whether this is fixed then there as well.

### DIFF
--- a/async/examples/dune
+++ b/async/examples/dune
@@ -1,11 +1,15 @@
 (executable
   (name    test_client)
   (modules test_client)
+  (public_name tls-test-client)
+  (package tls-async)
   (preprocess (pps ppx_jane))
   (libraries async core core_unix.command_unix tls-async))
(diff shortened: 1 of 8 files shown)


hannesm:
it's green, merging

PR #505, opened 2024-08-21
reduce the errors

hannesm:
There used to be an incomprehensible mess of errors. Now, those have been condensed to much fewer errors, all named as in the TLS RFCs.

Also reduce the variant of Packet.alert with only these that are actually used.


hannesm:
I started this quite some time ago, and finally got to finish it. This sometimes removes some information, sometimes it adds some. I haven't found any code that used the information in the errors, thus I'll push this through when CI agrees. Less code is better. Also TLS 1.3 has some guidelines on which alerts to send out.

PR #507, opened 2024-08-22
Fix the test on x86_32 architecture where our test is a bit too fast between the creation of the certificate and the authenticator called by the TLS handshake

dinosaure:
/cc @hannesm


hannesm:
thanks

PR #508, opened 2024-08-22
Fix file-descriptor leak into the miou impl. spotted by httpcats

dinosaure:
I spotted a file-descriptor leak on my `httpcats`'s test. The fix does not change the behavior of `tls-miou-unix` about `read` and `write` after a `shutdown` (the fuzzer should verify that) but ensure that we actually close the file-descriptor when we can.


hannesm:
thanks :) if you'd like a release, please let me know. maybe a release of only tls-miou-unix would be good?

dinosaure:
Yes, a release will be super useful :) . Not sure if it's easy to break the release cycle of all packages due to the `{= version}` constraint.

hannesm:
I can manually do a release where the tls-miou-unix.1.0.1 has a bound to tls =1.0.0...

PR #510, opened 2024-08-29
tls-lwt: read add an optional ?off argument

hannesm:
useful e.g. in albatross where we may have partial reads.

### DIFF
--- a/lwt/tls_lwt.ml
+++ b/lwt/tls_lwt.ml
@@ -153,12 +153,12 @@ module Unix = struct
             handle tls (String.sub (Bytes.unsafe_to_string t.recv_buf) 0 n)
           | `Closed -> Lwt.return `Eof
 
-  let rec read t buf =
+  let rec read t ?(off = 0) buf =
 
     let writeout res =
       let rlen = String.length res in
-      let n    = min (Bytes.length buf) rlen in
-      Bytes.blit_string res 0 buf 0 n ;
+      let n    = min (Bytes.length buf - off) rlen in
+      Bytes.blit_string res 0 buf off n ;
       t.linger <-
         (if n < rlen then Some (String.sub res n (rlen - n)) else None) ;
       Lwt.return n in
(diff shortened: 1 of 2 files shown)


lwt/tls_lwt.ml
```diff
@@ -153,12 +153,12 @@ module Unix = struct
             handle tls (String.sub (Bytes.unsafe_to_string t.recv_buf) 0 n)
           | `Closed -> Lwt.return `Eof
 
-  let rec read t buf =
+  let rec read t ?(off = 0) buf =
```
reynir, line 156: I guess we should add a check that `off` is non-negative and less than `Bytes.length buf`. Else we get an `Invalid_argument _` exception from `Bytes.blit_string` which is less clear.

lwt/tls_lwt.ml
```diff
@@ -181,7 +181,11 @@ module Unix = struct
             ( t.state <- inject_state tls t.state ; write_t t tlsdata )
         | None -> Lwt.reraise @@ Invalid_argument "tls: write: socket not ready"
 
-  let write t cs = writev t [cs]
+  let write t ?(off = 0) cs =
+    let buf =
+      if off = 0 then cs else String.sub cs off (String.length cs - off)
+    in
+    writev t [buf]
```
reynir, line 188: This makes a copy which you would not expect from the type :/ but maybe there is room for improvement later? I realize this probably goes  deep into the engine...

hannesm: yes, I disliked this as well. I can see two options: remove the `off`, or improve the performance later. I guess we'll need quite some adjustments to the engine...

lwt/tls_lwt.ml
hannesm: that's a great idea. I'll raise an exception..

hannesm: b9cb064

lwt/tls_lwt.ml
reynir: Maybe it's fine if we add a comment in the documentation that it currently makes a copy. Then, I wonder why not have a `?len` as well? I'm undecided.

hannesm: > Maybe it's fine if we add a comment in the documentation that it currently makes a copy. Then, I wonder why not have a `?len` as well? I'm undecided.

I find adding a documentation is not worth it. It will change. All the other functions as well don't have documentation about potential allocations.

I pushed c001750 which removed the `?off` from `write`. It makes the API very asymmetric (taking writev into account).

PR #511, opened 2024-09-29
Fix the recursion of Tls_miou_unix.read_in

dinosaure:
Spotted by caqti, this patch fix the recursion of read_in and keep the offset and the length calculated along the recursion even if, at the beginning, they are optional arguments.

### DIFF
--- a/miou/tls_miou_unix.ml
+++ b/miou/tls_miou_unix.ml
@@ -161,7 +161,7 @@ let rec read_in flow ?(off= 0) ?len buf =
   | Some res -> write_in res
   | None -> (
       match read_react flow with
-      | None -> read_in flow buf
+      | None -> read_in ~off ~len flow buf
       | Some res -> write_in res)
 
 let writev flow bufs =


miou/tests/fuzz.ml
```diff
@@ -38,6 +38,7 @@ module Ca = struct
       @ [ Relative_distinguished_name.singleton (CN "Ephemeral CA for fuzzer") ])
 
   let cacert_lifetime = Ptime.Span.v (365, 0L)
+  let cacert_serial_number = "\xde\xad\xbe\xef"
```
hannesm, line 41: is this needed?

miou/tls_miou_unix.ml
```diff
@@ -161,7 +161,7 @@ let rec read_in flow ?(off= 0) ?len buf =
   | Some res -> write_in res
   | None -> (
       match read_react flow with
-      | None -> read_in flow buf
+      | None -> read_in ~off ~len flow buf
```
hannesm, line 164: looks fine to me :)

hannesm:
when we get x509 into a nice shape, we won't need the fuzz changes, or do we? I'm happy to cut a release of the tls-miou-unix changes

miou/tests/fuzz.ml
dinosaure: We can revert my last commit if you want :+1:.

dinosaure:
> when we get x509 into a nice shape, we won't need the fuzz changes, or do we?

Yes, with the x509 fix, we will not need my last commit :+1:. We can revert it if you prefer.

hannesm:
yes, so I removed the latter commit. thanks a lot!

PR #515, opened 2025-04-13
allocate less: the case where huge data is sent via send_application_data

hannesm:
previously, we had a nearly exponential blowup at the border of 1 lsl 14 where the data was cut into pieces (but we still hold our hands at the old data)

### DIFF
--- a/lib/engine.ml
+++ b/lib/engine.ml
@@ -300,7 +300,7 @@ let encrypt_records encryptor version records =
           doit st ((ty, buf) :: acc) (off + len)
         else
           let st, ty, buf = encrypt version st ty buf off (bufl - off) in
-          st, List.rev ((ty, buf) :: acc)
+          st, (ty, buf) :: acc
       in
       let st, res =
         if String.length buf >= 1 lsl 14 then
--- a/lib/engine.ml
+++ b/lib/engine.ml
@@ -297,7 +297,7 @@ let encrypt_records encryptor version records =
         if bufl - off >= 1 lsl 14 then
           let len = 1 lsl 14 in
           let st, ty, buf = encrypt version st ty buf off len in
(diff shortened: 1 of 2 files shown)


lib/engine.ml
```diff
@@ -103,17 +105,16 @@ let encrypt (version : tls_version) (st : crypto_state) ty buf =
...
-          let signature = Crypto.mac mac mac_k pseudo_hdr buf in
-          buf ^ signature
+          Crypto.pseudo_header seq ty ver len
         in
         let c_st, enc =
           match ctx.cipher_st with
           | CBC c ->
              let enc iv =
-               let to_encrypt = to_encrypt c.hmac c.hmac_secret in
+               (* TODO only until digestig goes beyond 1.2.0 (feedable hmac) *)
```
reynir, line 114: Small typo:
```suggestion
               (* TODO only until digestif goes beyond 1.2.0 (feedable hmac) *)
```

lib/engine.ml
```diff
@@ -288,21 +289,31 @@ let rec separate_records : string -> ((tls_hdr * string) list * string, failure)
...
+          st, (ty, buf) :: acc
+      in
+      let st, res =
+        if String.length buf >= 1 lsl 14 then
+          doit st [] 0
+        else
+          let st, ty, buf =
+            encrypt version st ty buf 0 (String.length buf)
+          in
+          st, [ ty, buf ]
```
reynir, line 312: Isn't this the same as just calling `doit st [] 0`?

lib/engine.ml
hannesm: thanks.

lib/engine.ml
hannesm: yes indeed, fixed in 8ff2cb3

PR #516, opened 2025-06-18
put CERTIFICATE_UNKNOWN back

copy:
This alert is sent by chromium on self-signed certificates, so it probably makes sense to have a useful error message, as opposed to "TLS alert from peer: unknown 46"


hannesm:
thanks

PR #518, opened 2025-08-31
Add an unix implementation of TLS

kit-ty-kate:
This is an updated version of https://github.com/mirleft/ocaml-tls/pull/443 rebased on top of master.
I then took the current `tls-miou` implementation and replaced `Miou_unix` by `Unix`

### DIFF
--- a/unix/tls_unix.ml
+++ b/unix/tls_unix.ml
@@ -144,6 +144,35 @@ let read socket =
             drain_handshake flow
         | `Error e -> Error e
         | `Eof     -> Error Closed
+
+  let reneg ?authenticator ?acceptable_cas ?cert ?(drop = true) flow =
+    match flow.state with
+    | `Eof        -> Error Closed
+    | `Error e    -> Error e
+    | `Active tls ->
+        match Tls.Engine.reneg ?authenticator ?acceptable_cas ?cert tls with
+        | None             ->
+            (* XXX make this impossible to reach *)
+            invalid_arg "Renegotiation already in progress"
+        | Some (tls', buf) ->
+            if drop then flow.linger <- [] ;
+            flow.state <- `Active tls' ;
+            fully_write flow.flow buf |> fun _ ->
+            drain_handshake flow |> function
+            | Ok _         -> Ok ()
+            | Error _ as e -> e
+
+  let key_update ?request flow =
+    match flow.state with
+    | `Eof        -> Error Closed
+    | `Error e    -> Error e
+    | `Active tls ->
+      match Tls.Engine.key_update ?request tls with
+      | Error _ -> invalid_arg "Key update failed"
+      | Ok (tls', buf) ->
+        flow.state <- `Active tls' ;
+        fully_write flow.flow buf |> check_write flow
+
   let close flow =
     match flow.state with
     | `Active tls ->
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
I have a PoC PR using this in ocamldap: https://github.com/kit-ty-kate/ocamldap/pull/10
If anyone knows how to use the connect function correctly i'd be happy to take a hint

unix/tls_unix.ml
```diff
@@ -0,0 +1,337 @@
+let src = Logs.Src.create "tls-miou"
```
kit-ty-kate, line 1:
```suggestion
let src = Logs.Src.create "tls-unix"
```

unix/tls_unix.ml
```diff
@@ -0,0 +1,337 @@
...
+      buf = Bytes.make read_buffer_size '\000';
+      rd_closed = false;
+    }
+  in
+  drain_handshake tls_flow
+
+let write flow ?(off = 0) ?len str =
+  let len = Option.value ~default:(String.length str - off) len in
+  if off < 0 || len < 0 || off > String.length str - len
+  then invalid_arg "Tls_miou.write";
```
kit-ty-kate, line 281:
```suggestion
  then invalid_arg "Tls_unix.write";
```

unix/tls_unix.ml
```diff
@@ -0,0 +1,337 @@
...
+let rec really_read_go t off len buf =
+  let len' = read t buf ~off ~len in
+  if len' == 0 then raise End_of_file
+  else if len - len' > 0
+  then really_read_go t (off + len') (len - len') buf
+
+let really_read t ?(off= 0) ?len buf =
+  let len = Option.value ~default:(Bytes.length buf - off) len in
+  if off < 0 || len < 0 || off > Bytes.length buf - len
+  then invalid_arg "Tls_miou.really_read";
```
kit-ty-kate, line 300:
```suggestion
  then invalid_arg "Tls_unix.really_read";
```

unix/tls_unix.ml
```diff
@@ -0,0 +1,337 @@
...
+let connect authenticator (v, port) =
+  let conf =
+    match Tls.Config.client ~authenticator () with
+    | Ok config -> config
+    | Error `Msg msg -> Fmt.invalid_arg "Configuration failure: %s" msg
+  in
+  let addr = resolve v (string_of_int port) in
+  let fd =
+    match addr with
+    | Unix.ADDR_UNIX _ -> invalid_arg "Tls_miou.connect: Invalid UNIX socket"
```
kit-ty-kate, line 318:
```suggestion
    | Unix.ADDR_UNIX _ -> invalid_arg "Tls_unix.connect: Invalid UNIX socket"
```

hannesm:
> > Somehow, the really_read returns always end of file for me... not sure what the issue is there
> 
> Depending on the size of the buffer you give. Do you have a reproducible example?

same as above, my buf is a Bytes.create 4094. read returns around 2700 bytes if i recall correctly.

hannesm:
> Don't have strong opinion about that, I just need to keep into my mind that if we find a bug on `tls-unix`, it's probably the same on `tls-miou-unix`.

Indeed, I put a comment in both ml files.

PR #520, opened 2026-03-12
eio,lwt,mirage: fix drain_handshake on Read_closed/Write_closed states

samoht:
Note: I am not totally sure this is the right fix -- I have been seeing this while writing larger e2e tests using eio (but I noticed the other backends had the same issue). I can add a minimal test that reproduces this for eio if needed (but there is no exiring infrastucture in the repo to test this I think?)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+## Unreleased
+
+* tls-eio, tls-lwt, tls-mirage, tls-miou-unix, tls-unix: `drain_handshake`
+  now returns once the handshake completes, even if the peer has already
+  closed one half of the connection. It still raises `End_of_file` if the
+  peer closes before the handshake finishes (#520 @samoht)
+
 ## v2.0.4 (2026-03-10)
 
 * Fix exception in parse_change_cipher_spec on empty input (#519 @samoht)
--- a/eio/tls_eio.ml
+++ b/eio/tls_eio.ml
@@ -143,7 +143,8 @@ module Raw = struct
       | (Some cs, Some l) -> t.linger <- Some (Cstruct.append l cs)
     in
     match t.state with
-    | `Active tls when not (Tls.Engine.handshake_in_progress tls) ->
+    | `Active tls | `Read_closed tls | `Write_closed tls
+      when not (Tls.Engine.handshake_in_progress tls) ->
         t
     | _ ->
         let cs = read_react t in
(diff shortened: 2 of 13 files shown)


hannesm:
So, what should happen here in drain_handshake, if the flow has already been (partially) closed?

Your change adds some code for `` `Read_closed | `Write_closed `` -- but what should happen in the `` `Closed `` state? And how useful is such a flow to be exposed, which is (partially) closed?

For `` `Closed `` and `` `Read_closed ``, the `read_react` function returns `` `Eof `` -- which results in `End_of_file` -- now your PR changes the semantics for `` `Read_closed ``, why?

For `` `Write_closed ``, I guess your fix is fine -- when the handshake is complete, we can start reading from the flow.

NB: Your PR does only change some effectful layers -- is this intentional? Namely, async, unix, and miou-unix are not touched.

hannesm:
NOTE: async doesn't have the half-closed states (yet?)

hannesm:
Dear @samoht, would you mind to comment on my remarks 3 weeks ago, please? Thanks!

samoht:
Pushed two follow-up commits:
 
- lwt/tests/drain_handshake.ml — regression test that writes the full TLS-1.3 flight + close_notify in one channel write. On unfixed code raises End_of_file; with the fix it passes.
- miou + unix backends — same drain_handshake shape, same bug, same fix.

Async doesn't have the half-closed states (yet?), so the same fix doesn't apply directly. Happy to do it as a separate PR if you wan

samoht:
If you'd prefer a different shape:  e.g. a dedicated exception like Handshake_complete_peer_closed so the caller knows the handshake succeeded and then the connection ended, happy to do that instead. That avoids the awkwardness of returning a half-dead flow.

hannesm:
Thanks for your additional commits, Thomas.

The main questions are still open, it would be nice if you could reply to them:
- 1. Your change adds some code for `` `Read_closed | `Write_closed `` -- but what should happen in the `` `Closed `` state? And how useful is such a flow to be exposed, which is (partially) closed?
- 2. For `` `Closed `` and `` `Read_closed ``, the `read_react` function returns `` `Eof `` -- which results in ``End_of_file`` -- now your PR changes the semantics for `` `Read_closed `` (which now returns ok, and not an error), why?

What I observe is you have some desire to deal with half-closed TLS sessions, but could you describe a bit more in which context they are useful (apart from "writing larger e2e tests with eio")?

An approach would be: write a changes entry in the form what changes for an user of this API, and the reasoning behind it.

samoht:
I've rebased with a smaller normal `tls-lwt` regression test and a changelog entry.

The issue is that `drain_handshake` uses `End_of_file` for two different situations:

- peer close before handshake completion: constructor failure
- `close_notify` after handshake completion: TLS end-of-data

This PR makes the second case return a normal flow where `read` is returning `0` instead. Still not totally sure that's the right thing to be honnest, but using End_of_file for both sounds also wrong. What do you think? Maybe it also needs to be clarified in the docstrings.

hannesm:
Thanks, I reworded the documentation slightly, and squash-merged as 7cbe51c7306a1ce93d8978650f319f5437a1cb2b

PR #523, opened 2026-06-24
Similar to ?host, allow an ?ip in client_of_fd in all backends

hannesm:
Also provide Config.ip with that behaviour. Also, in backends where we resolve, if an IP address is provided, use that directly for validating the certificate to contain the IP address.


hannesm:
CI looks fine, merging.

PR #524, opened 2026-06-25
tls: keep the client role in the TLS 1.3 client KeyUpdate handler

samoht:
Fix a typo (Server13 -> Client13) in the client KeyUpdate handler. The impact is limited: the client is left in the server role, so a following NewSessionTicket is rejected and the connection gets stuck.


hannesm:
thanks

PR #527, opened 2026-07-19
Use warning name instead of number

hannesm:


hannesm:
CI error is that "core" or "core-unix" doesn't compile on riscv64.

PR #529, opened 2026-08-20
Let inject_state follow the engine's half-close

samoht:
The engine already knows whether a "close_notify" has been received or sent, in `State.t`'s `read_closed` and `write_closed`, but nothing exposes them. So each adapter derives the same two facts a second time: it turns `handle_tls`'s `` `Eof `` into a `half_close ... `read`, and follows its own `send_close_notify` with a `half_close ... `write`.

This PR exposes the two predicates and has `inject_state` apply them, so a state that carries an engine follows that engine's half-close instead of the caller having to remember to say so. That is what #521 and #522 were: a call site that put the engine back without carrying the half-close with it.

The adapters keep their own `half_close`, which still records the closes the engine cannot see, such as a transport EOF or a local `` `read `` shutdown. One commit per backend. tls-async has no such wrapper and is untouched.

### DIFF
--- a/eio/tls_eio.ml
+++ b/eio/tls_eio.ml
@@ -40,7 +40,10 @@ module Raw = struct
     | `Write_closed _, (`read | `read_write) -> `Closed
     | (`Closed | `Error _) as e, (`read | `write | `read_write) -> e
 
-  let inject_state tls = function
+  let inject_state tls state =
+    let state = if Tls.Engine.read_closed tls then half_close state `read else state in
+    let state = if Tls.Engine.write_closed tls then half_close state `write else state in
+    match state with
     | `Active _ -> `Active tls
     | `Read_closed _ -> `Read_closed tls
     | `Write_closed _ -> `Write_closed tls
@@ -62,10 +65,8 @@ module Raw = struct
 
     let handle tls buf =
       match Tls.Engine.handle_tls tls buf with
-      | Ok (state', eof, `Response resp, `Data data) ->
-          let state' = inject_state state' t.state in
-          let state' = Option.(value ~default:state' (map (fun `Eof -> half_close state' `read) eof)) in
-          t.state <- state' ;
+      | Ok (state', _eof, `Response resp, `Data data) ->
+          t.state <- inject_state state' t.state ;
           Option.iter (try_write_t t) resp;
           Option.map Cstruct.of_string data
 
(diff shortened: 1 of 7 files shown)


hannesm:
This may be a good idea in general (I'm still not sure about it, though), but would you please explain:
- did you write that code (& comments) yourself?
- why refer to 8446 sec 6.1, the very generic alert section -- how will this help anyone?
- why are there whitespace changes which pollute the diff?

samoht:
(this is still a WIP PR, hence the Draft status)

samoht:
I think this is ready for a first review. I think the only non-obvious part of the patch is that the current `handle_tls` loop does only this:

```ocaml
        if state.read_closed then begin
          Tracing.debug (fun m -> m "eof-out") ;
          Some `Eof
```

and never check `write_closed` so this PR is doing something more by ignoring the `Eof return and checking directly the state machine. Do you remember if that was expected? maybe the write_closed path is different? Or maybe it doesn't matter?

hannesm:
So, would you remind me: what is the issue you are facing, more concretely? And what is the solution you propose? I'm quite confused by reading your initial comment and then taking a brief look at the patch.

> I think this is ready for a first review. I think the only non-obvious part of the patch is that the current handle_tls loop does only this:
>
>        if state.read_closed then begin
>          Tracing.debug (fun m -> m "eof-out") ;
>          Some `Eof
>
> and never check write_closed

Wait a moment, what is `handle_tls` for? For handling incoming buffers on the TLS connection. Why should it ever do something with `write_closed`? If the read side has been closed, we don't read anything anymore. That's it. If the write side has been closed, we read something, but we never write anything.

Isn't this the whole story behind shutdown and half-close connections?

> so this PR is doing something more by ignoring the `Eof return and checking directly the state machine.

uhm, ok. I don't actually understand what you mean with this sentence.

> Do you remember if that was expected? maybe the write_closed path is different? Or maybe it doesn't matter?

As explained above, indeed half-closed situations are expected, and the TLS read part only cares if the read side is closed or not.

Before continuing with the review, I'm still wondering whether you could disclosure three things:
- your motivation (what is the issue you're facing)?
- whether you used any automated tools for developing the patch or the comments in this PR?
- what is based on the first question a good test case that failed before this patch, and works after this patch?

hannesm:
Dear Thomas, any updates on this PR? Any answers to my comment above?

samoht:
Sorry I unfortunately am running short of time. I'll close this PR and open it when I get more time to do this properly. Apology for the noise.

